PR #100, opened 2023-12-01
Reroute Refactor

### LINKED ISSUE #67: Logstash Ingest Pipelines - support for "reroute" processor
Elasticsearch 8.8 will include the [reroute](https://www.elastic.co/guide/en/elasticsearch/reference/8.8/reroute-processor.html) processor, that allows `to route a document to another target index or data stream`.
Example configuration:

```json
{
  "reroute": {
    "tag": "nginx",
    "if" : "ctx?.log?.file?.path?.contains('nginx')",
    "dataset": "nginx"
  }
}
```

Namely, regarding the execution of the pipelines, the documentation states:

> After a `reroute` processor has been executed, all the other processors of the current pipeline are skipped, including the final pipeline. If the current pipeline is executed in the context of a [Pipeline](https://www.elastic.co/guide/en/elasticsearch/reference/8.8/pipeline-processor.html), the calling pipeline will be skipped, too. This means that at most one reroute processor is ever executed within a pipeline, allowing to define mutually exclusive routing conditions, similar to a if, else-if, else-if, … condition.

yaauie:
Supercedes: https://github.com/elastic/logstash-filter-elastic_integration/pull/97

It looks like #97 lost its link to my fork when the main repo was made public. This PR continues the effort.

>  - Introduces `IntegrationRequest` and async-safety, to ensure that out `EventProcessor#processEvents` CANNOT return with async work from the batch still in-flight. An event is marked complete by using any of the `IngegrationRequest#complete` methods, which accept, transform, or replace the incoming event with the effective result.
>  - Introduces `IndexNameToPipelineNameResolver` which enables rerouted events to resolve their next pipeline name using only the index name.
>  - Introduces `EventToIndexNameResolver` which enables events to always have their resolved index name, which is necessary for the reroute processor's tracking of index history for cycle detection
>  - Breaks `DatastreamEventToPipelineNameResolverTest` implementation into separate `DatastreamEventToIndexNameResolver` and `ElasticsearchIndexNameToPipelineNameResolver`
>  - Rearranges `EventProcessor` to have injection points for re-routed pipelines to be run using the same execution code as first-pass pipelines
> 
> Resolves #67


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 0.0.4
   - Re-syncs with Elasticsearch 8.11 [#91](https://github.com/elastic/logstash-filter-elastic_integration/pull/91)
-  - Adds support for Geoip Processor to use databases from Logstash's Geoip Database Management service [#88](https://github.com/elastic/logstash-filter-elastic_integration/pull/88)
+  - Adds support for `reroute` processor [#97](https://github.com/elastic/logstash-filter-elastic_integration/issues/97)
```
yaauie, line 3:
```suggestion
  - Adds support for `reroute` processor [#100](https://github.com/elastic/logstash-filter-elastic_integration/pull/100)
```

src/main/java/co/elastic/logstash/filters/elasticintegration/IntegrationBatch.java
```diff
@@ -0,0 +1,51 @@
+package co.elastic.logstash.filters.elasticintegration;
```
yaauie, line 1:
```suggestion
/*
 * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V.
 * under one or more contributor license agreements. Licensed under the
 * Elastic License 2.0; you may not use this file except in compliance
 * with the Elastic License 2.0.
 */
package co.elastic.logstash.filters.elasticintegration;
```

mashhurs (approved):
LGTM!

Issue #99, opened 2023-11-29
Plugin doesn't connect to ES with Logstash encoded API key.


mashhurs:
### Description
When connecting to ES (confirmed on Serverless project), plugin needs to use API key. The plugin is happy with `Encoded` option, but when using the key from Logstash option, plugin raises `401 HTTP unauthorized` error. Since our rest of `es-input/output/filter` plugins use Logstash encoded key, we should either standardize it or document this case.

![image](https://github.com/elastic/logstash-filter-elastic_integration/assets/99575341/cf7feabd-4cac-42c3-933c-625f08006ff4)

yaauie:
Looking at the [Elasticsearch API Key docs]( https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html), we _severely_ overload the terms “API Key” vs `api_key` vs `ApiKey` in our stack.

An _Elasticsearch "API Key"_ is a conceptual collection of attributes:
 - `id`: a unique identifier for subsequent management of the generated api key
 - `name`: a user-friendly name that was provided while creating (or updating?) the api key
 - `expiration`: the expiration in UNIX-epoch millis if-and-only-if one was set when creating the "API Key"
 - `api_key`: a random string that is useless on its own
 - `encoded`: a useful string that can be used as an `Authorization: ApiKey ${encoded}` header to authenticate a request to Elasticsearch.

This plugin has an `api_key` that accepts _only_ the _Elasticsearch "API Key"_'s `encoded` form, which it includes verbatim in a `Authorization: ApiKey ${api_key}` header to authenticate to Elasticsearch. It uses a form that is both provided _by_ and acceptable _to_ Elasticsearch, without any extra steps. This makes sense to me. Things that are external to Elasticsearch should not need to guess their way through the details of that `encoded` value.

Logstash's other ES plugins that have an `api_key` option do _not_ accept the `encoded` form, but instead require that the user provide a string that _manually_ concatenates the _Elasticsearch "API Key"_'s `id` attribute, a literal colon (`:`), and its `api_key` attribute. This value is _technically_ a "decoded" form of the _Elasticsearch "API Key"_'s `encoded` attribute, and we provide helpers in the Kibana UI to make its generation less manual, but it is a representation of the key that is neither provided _by_ nor acceptable _to_ Elasticsearch.

[...]

PR #101, opened 2023-12-01
stub: intercept non-encoded `api_key` and encode them

yaauie:
Resolves: #99 

TODO: there are mocked responses from Serverless that include `base64` in the error response that need to be re-recorded. It is possible that we are relying on the previously-broken behaviour in our detection of Serverless, or merely that we need to validate behaviour in the presence of an `HTTP 401` and used an `HTTP 401` that included this message.


mashhurs (approved):
LGTM!

mashhurs (approved):

mashhurs:
I am going to go head and merge this change since I have _only_ added unit test cases.

PR #93, opened 2023-11-13
Plugin build pipeline

mashhurs:
### Description

A plugin build pipeline which builds the plugin on various ES versions including snapshots and main branch to detect the issues earlies.

Tasks considered:
- [x] Build on latest release (dynamically resolve 8.11) and run integration tests with release version (same 8.11) docker
- [x] Build on latest snapshot (dynamically resolve 8.11.x-snapshot) and run integration tests with snapshot version (same 8.11.x-snapshot) docker
  - THIS IS IMPOSSIBLE NOW SINCE [OUR RELEASE DEFINITION](https://raw.githubusercontent.com/elastic/logstash/main/ci/logstash_releases.json) DOESN'T CARRY DETAILS
- [x] Build on main and run integration tests with main snapshot version (8.12 for this case) docker
- [x] A definition to send slack notification when fails

PR #103, opened 2023-12-01
Create build plugin pipeline

mashhurs:
### Description
**Replacing #93 since repo went public, I had to re-open with a new PR.**

A plugin build pipeline which builds the plugin on various ES versions including snapshots and main branch to detect the issues earlies.

Tasks considered:
- [x] Build on latest release (dynamically resolve 8.11) and run integration tests with release version (same 8.11) docker
- [x] Build on latest snapshot (dynamically resolve 8.11.x-snapshot) and run integration tests with snapshot version (same 8.11.x-snapshot) docker
  - THIS IS IMPOSSIBLE NOW SINCE [OUR RELEASE DEFINITION](https://raw.githubusercontent.com/elastic/logstash/main/ci/logstash_releases.json) DOESN'T CARRY DETAILS
- [x] Build on main and run integration tests with main snapshot version (8.12 for this case) docker
- [x] A definition to send slack notification when fails


yaauie (commented):
This is worth moving forward with, even if I have additional thoughts. Feel free to spin them off as a secondary feature-request.

Ideally, we would test daily against:
 - the most recent release in the 8.x series (✅ maps to `ELASTIC_STACK_VERSION=8.x SNAPSHOT=false`)
 -  a snapshot of the _next_ release in the 8.x series (❌ no current mapping, since `logstash_releases.json` has no entry representing `8.x` that has artifacts built from `8.11`, which as of this writing is where the next release will come from)
 - the _future_ release from `main` (🟡 currently indirectly satisfied by `ELASTIC_STACK_VERSION=8.x SNAPSHOT=true`, since we maintain outside knowledge that `8.12.0-SNAPSHOT` is built from `main`)

With these changes, we get improvements, but I think we need to add an entry to the upstream `logstash_releases.json` to get what we are looking for.

I'm expressing in YAML because I need comments and clean diffs, but the source file of course is actually JSON.

As-implemented now, `snapshots[8.x]` points to an artifact built off of the `main` branch;
once a feature branch is cut, we no longer point to it. `snapshots[8.x]` should be renamed to `snapshots[main]`, and we should introduce a _new_ `snapshots[8.x]` that points to a snapshot built off the latest 8.x branch:

~~~ diff
 snapshots:
   "7.x": "7.17.16-SNAPSHOT" # next release in 7.x series
-  "8.x": "8.12.0-SNAPSHOT" # future minor (pre-ff)
+  "8.x": "8.11.2-SNAPSHOT" # next release in 8.x series
+  "main": "8.12.0-SNAPSHOT" # future minor (pre-ff)
~~~


When 8.12 feature-freezes, the diff of that FF would look like:

~~~ diff
 snapshots:
   "7.x": "7.17.16-SNAPSHOT" # next release
-  "8.x": "8.11.2-SNAPSHOT"  # next release in 8.x series
+  "8.x": "8.12.0-SNAPSHOT"  # next release in 8.x series
-  "main": "8.12.0-SNAPSHOT" # future minor (pre-ff)
+  "main": "8.13.0-SNAPSHOT" # future minor (pre-ff)
~~~

When 8.11.2 drops the diff would look like:

~~~ diff
 releases:
   "5.x": "5.6.16"
   "6.x": "6.8.23"
   "7.x": "7.17.15" # latest minor, latest patch
-  "8.x": "8.11.1" # latest minor, latest patch
+  "8.x": "8.11.2" # latest minor, latest patch
 snapshots:
   "7.x": "7.17.16-SNAPSHOT" # next release
-  "8.x": "8.11.2-SNAPSHOT"  # next release in 8.x series
+  "8.x": "8.11.3-SNAPSHOT"  # next release in 8.x series
   "main": "8.12.0-SNAPSHOT" # future minor (pre-ff)
~~~

When 8.12.0 drops the diff would look like:

~~~ diff
 releases:
   "5.x": "5.6.16"
   "6.x": "6.8.23"
   "7.x": "7.17.15" # latest minor, latest patch
-  "8.x": "8.11.2" # latest minor, latest patch
+  "8.x": "8.12.0" # latest minor, latest patch
 snapshots:
   "7.x": "7.17.16-SNAPSHOT" # next release
-  "8.x": "8.11.3-SNAPSHOT"  # next release in 8.x series
+  "8.x": "8.12.1-SNAPSHOT"  # next release in 8.x series
   "main": "8.13.0-SNAPSHOT" # future minor (pre-ff)
~~~

This would give us the ability we need, to
 - test against the latest release `ELASTIC_STACK_VERSION=8.x SNAPSHOT=false`
 - test against the next release `ELASTIC_STACK_VERSION=8.x SNAPSHOT=true`
 - test against upcoming changes pre-FF `ELASTIC_STACK_VERSION=main SNAPSHOT=true`

mashhurs:
> This is worth moving forward with, even if I have additional thoughts. Feel free to spin them off as a secondary feature-request.
> 
> Ideally, we would test daily against:
> 
> * the most recent release in the 8.x series (✅ maps to `ELASTIC_STACK_VERSION=8.x SNAPSHOT=false`)
> * a snapshot of the _next_ release in the 8.x series (❌ no current mapping, since `logstash_releases.json` has no entry representing `8.x` that has artifacts built from `8.11`, which as of this writing is where the next release will come from)
> * the _future_ release from `main` (🟡 currently indirectly satisfied by `ELASTIC_STACK_VERSION=8.x SNAPSHOT=true`, since we maintain outside knowledge that `8.12.0-SNAPSHOT` is built from `main`)
> 
> With these changes, we get improvements, but I think we need to add an entry to the upstream `logstash_releases.json` to get what we are looking for.
> 
> I'm expressing in YAML because I need comments and clean diffs, but the source file of course is actually JSON.
> 
> As-implemented now, `snapshots[8.x]` points to an artifact built off of the `main` branch; once a feature branch is cut, we no longer point to it. `snapshots[8.x]` should be renamed to `snapshots[main]`, and we should introduce a _new_ `snapshots[8.x]` that points to a snapshot built off the latest 8.x branch:
> 
> ```diff
>  snapshots:
>    "7.x": "7.17.16-SNAPSHOT" # next release in 7.x series
> -  "8.x": "8.12.0-SNAPSHOT" # future minor (pre-ff)
> +  "8.x": "8.11.2-SNAPSHOT" # next release in 8.x series
> +  "main": "8.12.0-SNAPSHOT" # future minor (pre-ff)
> ```
> 
> When 8.12 feature-freezes, the diff of that FF would look like:
> 
> ```diff
>  snapshots:
>    "7.x": "7.17.16-SNAPSHOT" # next release
> -  "8.x": "8.11.2-SNAPSHOT"  # next release in 8.x series
> +  "8.x": "8.12.0-SNAPSHOT"  # next release in 8.x series
> -  "main": "8.12.0-SNAPSHOT" # future minor (pre-ff)
> +  "main": "8.13.0-SNAPSHOT" # future minor (pre-ff)
> ```
> 
> When 8.11.2 drops the diff would look like:
> 
> ```diff
>  releases:
>    "5.x": "5.6.16"
>    "6.x": "6.8.23"
>    "7.x": "7.17.15" # latest minor, latest patch
> -  "8.x": "8.11.1" # latest minor, latest patch
> +  "8.x": "8.11.2" # latest minor, latest patch
>  snapshots:
>    "7.x": "7.17.16-SNAPSHOT" # next release
> -  "8.x": "8.11.2-SNAPSHOT"  # next release in 8.x series
> +  "8.x": "8.11.3-SNAPSHOT"  # next release in 8.x series
>    "main": "8.12.0-SNAPSHOT" # future minor (pre-ff)
> ```
> 
> When 8.12.0 drops the diff would look like:
> 
> ```diff
>  releases:
>    "5.x": "5.6.16"
>    "6.x": "6.8.23"
>    "7.x": "7.17.15" # latest minor, latest patch
> -  "8.x": "8.11.2" # latest minor, latest patch
> +  "8.x": "8.12.0" # latest minor, latest patch
>  snapshots:
>    "7.x": "7.17.16-SNAPSHOT" # next release
> -  "8.x": "8.11.3-SNAPSHOT"  # next release in 8.x series
> +  "8.x": "8.12.1-SNAPSHOT"  # next release in 8.x series
>    "main": "8.13.0-SNAPSHOT" # future minor (pre-ff)
> ```
> 
> This would give us the ability we need, to
> 
> * test against the latest release `ELASTIC_STACK_VERSION=8.x SNAPSHOT=false`
> * test against the next release `ELASTIC_STACK_VERSION=8.x SNAPSHOT=true`
> * test against upcoming changes pre-FF `ELASTIC_STACK_VERSION=main SNAPSHOT=true`

This is a wrap up from our offline discussion and [the change to `release_defition.json`](https://github.com/elastic/logstash/pull/15631) introduces `main` and updates next minor `8.11` release (current snapshot). Once sorted out, I am planning apply changes to our release process and share with the team.

catalog-info.yaml
```diff
@@ -1,9 +1,13 @@
+# ***********************************
+# SECTION START: Backstage logstash-filter-elastic-integration component
+# ***********************************
+
 ---
-# yaml-language-server: $schema=https://gist.githubusercontent.com/elasticmachine/988b80dae436cafea07d9a4a460a011d/raw/rre.schema.json
+# yaml-language-server: $schema=https://gist.githubusercontent.com/elasticmachine/988b80dae436cafea07d9a4a460a011d/raw/e57ee3bed7a6f73077a3f55a38e76e40ec87a7cf/rre.schema.json
```
dliappis, line 6: We should be using the latest version of the RRE schema, as it was before: https://gist.githubusercontent.com/elasticmachine/988b80dae436cafea07d9a4a460a011d/raw/rre.schema.json so that it reflects the latest changes.
For more details see (the internal link) https://elastic.slack.com/archives/C0522G6FBNE/p1700477179859799 and https://elasticco.atlassian.net/browse/CI-322?focusedCommentId=96184

catalog-info.yaml
```diff
@@ -18,19 +22,97 @@ metadata:
...
+      pipeline_file: .buildkite/pull-request-pipeline.yml
       repository: elastic/logstash-filter-elastic_integration
+      provider_settings:
+        build_pull_requests: true
+        build_pull_request_forks: false
+        build_branches: false
+        build_tags: false
+        filter_enabled: true
+        filter_condition: >-
+          build.creator.name == 'elasticmachine' && build.pull_request.id != null
```
dliappis, line 64: In a pull request CI workflow, if the user pushes new commits to the branch, usually we want to cancel any existing or queued runs. This can be done by adding:

```
     cancel_intermediate_builds: true
      skip_intermediate_builds: true
```

in `provider_settings_settings:`. Docs in https://buildkite.com/docs/pipelines/skipping#skip-queued-intermediate-builds / https://buildkite.com/docs/pipelines/skipping#cancel-running-intermediate-builds

catalog-info.yaml
```diff
@@ -18,19 +22,97 @@ metadata:
...
+        ELASTIC_PR_COMMENTS_ENABLED: 'true'
+        ELASTIC_SLACK_NOTIFICATIONS_ENABLED: 'true'
+        SLACK_NOTIFICATIONS_CHANNEL: '#logstash-build'
+        SLACK_NOTIFICATIONS_ON_SUCCESS: 'false'
       teams:
         ingest-fp:
-          access_level: BUILD_AND_READ
-  owner: group:ingestion-team
+          access_level: MANAGE_BUILD_AND_READ
+        everyone:
```
dliappis, line 73: @roaksoax does it make sense to add this job under the [logstash buildkite organization](https://buildkite.com/elastic?team=logstash)? If yes, we should add here:

```
        logstash:
          access_level: MANAGE_BUILD_AND_READ
```

While at it, let's explicitly add rights to the `ingest-eng-prod` org too with:

```
        ingest-eng-prod:
          access_level: MANAGE_BUILD_AND_READ
```

catalog-info.yaml
```diff
@@ -18,19 +22,97 @@ metadata:
...
+      schedules:
+        Daily Plugin Build:
+          branch: main
+          cronline: "@daily"
+          message: "Run plugin build every day."
+      teams:
+        ingest-fp:
+          access_level: MANAGE_BUILD_AND_READ
+        everyone:
+          access_level: READ_ONLY
```
dliappis, line 118: same comment as above here

.buildkite/scripts/resolve_es_branch.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -e
+
+VERSION_URL="https://raw.githubusercontent.com/elastic/logstash/main/ci/logstash_releases.json"
+
+echo "Fetching versions from $VERSION_URL"
+VERSIONS=$(curl --silent $VERSION_URL)
```
dliappis, line 8: I usually like to protect CI jobs from transient network failures. There are two avenues here:

1. Use curl's native `--retry`/`--retry-delay` e.g. `VERSIONS=$(curl --retry 5 --retry-delay 5 -fsSL $VERSION_URL)` (`S` -> show error messages if request fails, in spite of `s` -> silent ; `L` follow redirects, `f` returns an error (thus ending the script if the HTTP response status in an error code)

2. Alternatively, if instead of `family/core-ubuntu-2204` you use one of the custom logstash CI vm images like `family/platform-ingest-logstash-ubuntu-2204`, you can take advantage of the pre-installed `retry` command with something like `VERSIONS=$(retry -t 5 -- curl -fsSL $VERSION_URL)`.
3.

.buildkite/scripts/resolve_es_branch.sh
```diff
@@ -0,0 +1,24 @@
...
+
+if [[ "$SNAPSHOT" == "true" ]]; then
+  key=$(echo "$VERSIONS" | jq '.snapshots."'"$ELASTIC_STACK_VERSION"'"')
```
dliappis, line 11: if you use `-r` with `jq` like: `jq -r '.snapshots."'"$ELASTIC_STACK_VERSION"'"')` you won't need to remove wrapped quotation marks later:

```
$ key=$(echo $VERSION_URL | jq '.snapshots."'"8.x"'"')
$ echo $key
"8.12.0-SNAPSHOT"
```

versus

```
$ key=$(echo $VERSION_URL | jq -r '.snapshots."'"8.x"'"')
$ echo $key
8.12.0-SNAPSHOT
```

.buildkite/scripts/resolve_es_branch.sh
```diff
@@ -0,0 +1,24 @@
...
+else
+  key=$(echo "$VERSIONS" | jq '.releases."'"$ELASTIC_STACK_VERSION"'"')
```
dliappis, line 13: same like above use `-r` with `jq` for raw output without `""`

.buildkite/scripts/resolve_es_branch.sh
```diff
@@ -0,0 +1,24 @@
...
+fi
+
+# remove the wrapped quotation marks ("")
+resolved_stack_version=$(echo "$key" | sed 's/\"//g')
```
dliappis, line 17: as mentioned above, not needed, if `-r` is used with `jq`

.buildkite/scripts/resolve_es_branch.sh
```diff
@@ -0,0 +1,24 @@
...
+
+IFS='.'
+read -a versions <<< "$resolved_stack_version"
+target_branch=${versions[0]}.${versions[1]}
```
dliappis, line 21: getting the major and minor version sounds like a good candidate for a [bash PE expression](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) i.e. replace all the three lines above with:

`target_branch=${resolved_stack_version%.*}`

this works from right to left and removes every character up to and including the first found `.`:

```
$ echo $key
8.12.0-SNAPSHOT

$ echo ${key%.*}
8.12
```

.buildkite/scripts/run_tests.sh
```diff
@@ -1 +1,8 @@
+if [ -z "$ELASTICSEARCH_TREEISH" ]; then
+  source .buildkite/scripts/resolve_es_branch.sh
+  echo "Resolved ELASTICSEARCH_TREEISH: ${ELASTICSEARCH_TREEISH}"
+else
+  echo "ELASTICSEARCH_TREEISH ${ELASTICSEARCH_TREEISH} is defined."
+fi
+
 mkdir -p .ci && curl -sL https://github.com/logstash-plugins/.ci/archive/buildkite-1.x.tar.gz | tar zxvf - --skip-old-files --strip-components=1 -C .ci --wildcards '*Dockerfile*' '*docker*' '*.sh' && .ci/docker-setup.sh && .ci/docker-run.sh
```
dliappis, line 8: same comment here for adding some retries to `curl` with `--retry` / `--retry-delay`

catalog-info.yaml
```diff
@@ -18,19 +22,97 @@ metadata:
...
     metadata:
-      description: ':logstash: The logstash-filter-elastic_integration :pipeline:'
-      name: logstash-filter-elastic_integration-ci
+      description: ':logstash: The logstash-filter-elastic_integration-pull-request :pipeline:'
+      name: logstash-filter-elastic_integration-pull-request
     spec:
-      pipeline_file: .buildkite/pipeline.yml
-      provider_settings:
-        build_pull_request_forks: true
+      pipeline_file: .buildkite/pull-request-pipeline.yml
```
dliappis, line 55: let's add a maximum timeout time, otherwise if it hangs, it can stay stuck up to 24hrs which is the default. Timeout can be defined at the `implementation.spec:` level, with:

```yaml
spec:
  implementation:
    spec:
      pipeline_file: ...
      maximum_timeout_in_minutes: 120
```

catalog-info.yaml
```diff
@@ -18,19 +22,97 @@ metadata:
...
   type: buildkite-pipeline
+  system: buildkite
+  implementation:
+    apiVersion: buildkite.elastic.dev/v1
+    kind: Pipeline
+    metadata:
+      description: ':logstash: The logstash-filter-elastic_integration-build :pipeline:'
+      name: logstash-filter-elastic_integration-build
+    spec:
+      pipeline_file: .buildkite/build-pipeline.yml
```
dliappis, line 102: same comment here like the previous pipeline definition, let's add a `maximum_timeout_in_minutes` as well here.

dliappis (commented):
Thank you for this work!

I did a pass and left some, mostly improvement related, comments.

.buildkite/scripts/resolve_es_branch.sh
mashhurs: Thanks so much for great feedback.
I have addressed and resolved _obvious_ changes.

- as we are using `family/core-ubuntu-2204` VM image (as discussed before need to spin up docker in CI), I have preferred 2nd option of suggestions;

catalog-info.yaml
mashhurs: adding this to `logstash` org under BK sounds good to me too.

.buildkite/scripts/resolve_es_treeish.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -e
```
dliappis, line 3: I suggest we switch this to `set -euo pipefail`.

- `-u`: will fail fast is a var is undefined e.g. `ELASTIC_STACK_VERSION` (I was caught by this while running the script in isolated fashion)
- `-o pipefail`: since we are using pipes e.g. `| jq -r ...` this will catch errors with `jq` and fail fast.
-

.buildkite/scripts/resolve_es_branch.sh
dliappis: > Thanks so much for great feedback. I have addressed and resolved _obvious_ changes.
> 
> * as we are using `family/core-ubuntu-2204` VM image (as discussed before need to spin up docker in CI), I have preferred 2nd option of suggestions;

You are welcome! Seems to be based on the [latest commit](https://github.com/elastic/logstash-filter-elastic_integration/pull/103/commits/afa65dc41ea10e4678fa2aaf1361f585eeee8a52) that you actually went with option 1. No problems at all there, just wanted to make sure we are on the same page.

dliappis (approved):
LGTM (provided the Buildkite job becomes green)

mashhurs:
> LGTM (provided the Buildkite job becomes green)

To provide more details, the pipeline becomes red actually and it is an expectation. The plugin is _not_ synced with ES `main` branch yet, and [it is planned](https://github.com/elastic/logstash-filter-elastic_integration/issues/107).

roaksoax:
Given that we need to create a build to test against 8.11, I think we should go ahead and land it as is. The failures on 8.12/main is because of the missing 8.12 ES snapshot. So lets land, and work next week to unblock this bui8ld on main/8.12

dliappis (approved):
LGTM with the latest changes

PR #112, opened 2023-12-07
Fix: when snapshot env var is undefined, script crashes.

mashhurs:
### Description
We define `SNAPSHOT` vars under pipeline env and when it is not defined recently introduced `resolve_es_treeish.sh` script crashes at [this point](https://github.com/elastic/logstash-filter-elastic_integration/blob/main/.buildkite/scripts/resolve_es_treeish.sh#L10). This PR sets default value to `SNAPSHOT` when undefined and allows script to run.

### DIFF
--- a/.buildkite/scripts/resolve_es_treeish.sh
+++ b/.buildkite/scripts/resolve_es_treeish.sh
@@ -7,7 +7,9 @@ VERSION_URL="https://raw.githubusercontent.com/elastic/logstash/main/ci/logstash
 echo "Fetching versions from $VERSION_URL"
 VERSIONS=$(curl --retry 5 --retry-delay 5 -fsSL $VERSION_URL)
 
-if [[ "$SNAPSHOT" == "true" ]]; then
+snapshot=${SNAPSHOT:-false}
+
+if [[ "$snapshot" == "true" ]]; then
   key=$(echo "$VERSIONS" | jq -r '.snapshots."'"$ELASTIC_STACK_VERSION"'"')
 else
   key=$(echo "$VERSIONS" | jq -r '.releases."'"$ELASTIC_STACK_VERSION"'"')
--- a/.buildkite/scripts/resolve_es_treeish.sh
+++ b/.buildkite/scripts/resolve_es_treeish.sh
@@ -1,15 +1,13 @@
 #!/bin/bash
 
-set -euo pipefail
+set +u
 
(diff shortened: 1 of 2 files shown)


catalog-info.yaml
```diff
@@ -67,9 +67,6 @@ spec:
       skip_intermediate_builds: true
       env:
         ELASTIC_PR_COMMENTS_ENABLED: 'true'
-        ELASTIC_SLACK_NOTIFICATIONS_ENABLED: 'true'
```
mashhurs, line 70: review note: I don't think we need to monitor PR failures.

.buildkite/scripts/resolve_es_treeish.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -euo pipefail
+set +u
```
dliappis, line 3: Do we need to cancel failing on errors entirely, which is what this does?

My understanding of the problem is that if `SNAPSHOT` is undefined we fail here: https://github.com/elastic/logstash-filter-elastic_integration/blob/main/.buildkite/scripts/resolve_es_treeish.sh#L10

So what if we just either:

1) change to `set -eo pipefail`, which shouldn't cause failures anymore if undefined variables are used?

2) right before https://github.com/elastic/logstash-filter-elastic_integration/blob/main/.buildkite/scripts/resolve_es_treeish.sh#L10 we `set +u`?

dliappis (commented):
Left a question

.buildkite/scripts/resolve_es_treeish.sh
mashhurs: Our expectation is `SNAPSHOT=true` is explicitly set if we need to test against snapshot, otherwise, we use release version. That is said, in order to test against release versions, we don't need to set `SNAPSHOT=false`.

dliappis: Ok, so in this case, we can simply skip failing hard if `SNAPSHOT` is undefined, so I think it's enough to `set +u` (or `set +o nounset`, if you prefer the verbose version [man page](https://www.gnu.org/software/bash/manual/bash.html#The-Set-Builtin)) just before the conditional. If we do that, `set -euo pipefail` can remain at the top of the file and benefit us from the usual error checks for any other potential problem.

dliappis (approved):
Lgtm

PR #114, opened 2023-12-19
Updating JAR output path to avoid using long path.

mashhurs:
## Description
[Embedding the plugin into core](https://buildkite.com/elastic/logstash-pull-request-pipeline/builds/514#018c6397-7760-4fb1-a1d4-34680c6a1e7a) and faced an issue that jar path exceeds path limit or [TarWriter we used when assembling](https://github.com/mashhurs/logstash/blob/69fbf46f0526b092aa17b46acfa82a29bc56bcce/rakelib/artifacts.rake#L469).
With this change, we will shorten the JAR path.

### BEFORE
268 chars: logstash-8.13.0-SNAPSHOT/vendor/bundle/jruby/3.1.0/gems/logstash-filter-elastic_integration-0.1.2-java/vendor/jar-dependencies/co/elastic/logstash/plugins/filter/elasticintegration/logstash-filter-elastic_integration/0.1.2/logstash-filter-elastic_integration-0.1.2.jar

### AFTER
214 chars: logstash-8.13.0-SNAPSHOT/vendor/bundle/jruby/3.1.0/gems/logstash-filter-elastic_integration-0.1.2-java/vendor/jar-dependencies/logstash-filter-elastic_integration/0.1.2/logstash-filter-elastic_integration-0.1.2.jar


yaauie:
The actual changes here LGTM, but if we were to simply add the jar to the `vendor` directory without any superfluous nesting we may be able to simply use JRuby's `Kernel#require_relative`, which would shorten the jar's path to:

155 chars: `logstash-8.13.0-SNAPSHOT/vendor/bundle/jruby/3.1.0/gems/logstash-filter-elastic_integration-0.1.2-java/vendor/logstash-filter-elastic_integration-0.1.2.jar`

mashhurs:
> The actual changes here LGTM, but if we were to simply add the jar to the `vendor` directory without any superfluous nesting we may be able to simply use JRuby's `Kernel#require_relative`, which would shorten the jar's path to:
> 
> 155 chars: `logstash-8.13.0-SNAPSHOT/vendor/bundle/jruby/3.1.0/gems/logstash-filter-elastic_integration-0.1.2-java/vendor/logstash-filter-elastic_integration-0.1.2.jar`

-  if we were to simply add the jar to the `vendor`: if we directly place under `vendor`, the `require_jar` method (in `jar_dependencies.rb`) requires three params (group id, artifact id and version). So, we need at least one path for group id. Maybe we can use `elastic` only? Like elastic - project - version as we organize packaging? How do you think?
  - in a reality, there is not issue in most OS file systems if path is over 256 limit. Pushing `TarWriter` would make more sense. Unfortunately, there is an open issue lasting over years for this.
- we may be able to simply use JRuby's `Kernel#require_relative`: ah this one I couldn't get. Do you mean instead `TarWriter` we use `Kernel#require_relative` first and zip?

build.gradle
```diff
@@ -427,9 +427,8 @@ tasks.withType(Test) {
 task vendorShadowJar(dependsOn: shadowJar) {
     description "vendors jar dependencies"
 
-    String vendorPathPrefix = "vendor/jar-dependencies"
-    String projectGroupPath = project.group.replaceAll('\\.', '/')
-    File projectJarFile = file("${vendorPathPrefix}/${projectGroupPath}/${project.name}/${project.version}/${project.name}-${project.version}.jar")
+    String vendorPathPrefix = "vendor/jar-dependencies/co/elastic"
```
mashhurs, line 430: review note: made a minimal change to shorten the path. We may directly place into vendor but we are using `jar-dependencies` (in gemspec file) as a required path and include jars as a standard in other Java based plugins.

jsvd:
the minimal change LGTM too.
I believe @yaauie's suggestion is to replace use of `require_jar` from jar_dependencies with a plain require or require_relative with the path to the jar, something like (pseudocode)

```groovy
        jarRequiresFile.withWriter { w ->
            w << "# AUTOGENERATED BY THE GRADLE SCRIPT. DO NOT EDIT.\n\n"
            w << '########################################################################\n' +
                 '# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V.\n' +
                 '# under one or more contributor license agreements. Licensed under the\n' +
                 '# Elastic License 2.0; you may not use this file except in compliance\n' +
                 '# with the Elastic License 2.0.\n' +
                 '########################################################################\n'
            w << '\n'
            w << "require_relative(::File.join("..", "vendor", "#{project.name}-#{project.version}.jar")\n"
        }
```

Issue #82, opened 2023-08-08
Doc: Fix asciidoc formatting and references so that we can pass docs-ci


karenzone:
Current docs source contains formatting errors that prevent a successful local build. 

The approach to [setting variables](https://github.com/elastic/logstash-filter-elastic_integration/blame/main/docs/index.asciidoc#L290-294) is not working in my local builds.  As a quick and dirty test, I removed the pass blocks from the variables, and the content built without errors.

PR #115, opened 2023-12-19
Doc: Formatting fixes

karenzone:
Fixes: #82


docs/index.asciidoc
```diff
@@ -54,7 +54,7 @@ Before upgrading this plugin or Logstash itself, please pay special attention to
 
 You will need to configure this plugin to connect to {es}, and may need to also need to provide local GeoIp databases.
 
-[source]
+[source,ruby]
```
karenzone, line 57: Required to make code sample render correctly

docs/index.asciidoc
```diff
@@ -292,8 +292,8 @@ This plugin supports the following configuration options plus the <<plugins-{typ
 
 // Variables for re-use in per-option docs
 :prohibit-ssl-disabled-effective: Cannot be combined with configurations that disable SSL
-:prohibit-ssl-disabled-explicit: Cannot be combined with <<plugins-{type}s-{plugin}-ssl_enabled, `ssl_enabled \=> false`>>.
-:prohibit-ssl-verify-none: Cannot be combined with <<plugins-{type}s-{plugin}-ssl_verification_mode, `ssl_verification_mode \=> none`>>.
+:prohibit-ssl-disabled-explicit: Cannot be combined with `<<plugins-{type}s-{plugin}-ssl_enabled>>=>false`.
+:prohibit-ssl-verify-none: Cannot be combined with `<<plugins-{type}s-{plugin}-ssl_verification_mode>>=>none`.
```
karenzone, line 296: Reformatting to make link text literal

karenzone:
@mashhurs The extra `+` signs in source are rendering literally throughout the page.  Here's an example: 

Screen Shot 2023-12-19 at 5 27 24 PM


I'm guessing that you want these entries to be literal?  For example (borrowed from screenshot): `_none`, `pipeline => "none"`, `_ingest_pipeline_failure`.  Is that correct? If so, I'll fix those instances as part of this PR.

docs/index.asciidoc
karenzone: This approach works, but I'm not convinced that it's best practice.  This approach doesn't save us much time or repetition, yet puts added burden on a user looking at docs in the plugin repo. 

Screen Shot 2023-12-19 at 5 35 53 PM

Rather than clearly stating the information with the setting, we're giving users a job to do in order to go find where we set up these variables.

karenzone: I'd like to get the initial docs published soon, and I'm willing to table this for a later discussion outside of this PR.

karenzone:
@mashhurs, please take a look and share your thoughts.  When we're done, I'll bump versions and publish to get these docs building for the [Logstash Reference](https://www.elastic.co/guide/en/logstash/current/filter-plugins.html).

mashhurs:
> @mashhurs The extra `+` signs in source are rendering literally throughout the page. Here's an example:
> 
> Screen Shot 2023-12-19 at 5 27 24 PM
> I'm guessing that you want these entries to be literal? For example (borrowed from screenshot): `_none`, `pipeline => "none"`, `_ingest_pipeline_failure`. Is that correct? If so, I'll fix those instances as part of this PR.

That's correct @karenzone ! Thank you for catching up.

karenzone:
Thanks for the confirmation, @mashhurs. I'll fix it tomorrow.

mashhurs (dismissed):
Thank you for the fixes! LGTM.

karenzone:
@mashhurs, new commits caused your review to be dismissed. Can I get a fresh LGTM, please?

mashhurs (approved):
Thank you for the fixes. LGTM!

karenzone:
Published at [v.0.1.3](https://rubygems.org/gems/logstash-filter-elastic_integration/versions/0.1.3-java)

PR #122, opened 2024-01-28
Basic E2E tests for the plugin.

mashhurs:
### Description
Adds integration tests for `["apache", "apache_tomcat", "elastic_agent", "elasticsearch", "m365_defender", "nginx", "tomcat"]` integrations.

### How to test
- Follow the `.buildkite/scripts/e2e/README.md` file

### Integrate with BK CI
- When calling script from the BK pipeline, we need to first resolve `ELASTIC_STACK_VERSION` and export it.
- Need prerequisites installed before running execution, `pip install -r .buildkite/scripts/e2e/requirements.txt`
- Run tests with `python .buildkite/scripts/e2e/main.py`


.buildkite/scripts/e2e/bootstrap.py
```diff
@@ -0,0 +1,177 @@
...
+                            f"output: {exec_result.output.decode('utf-8')}.")
+        print("Plugin installed successfully.")
+
+    def __reload_container(self):
+        result = subprocess.run(["docker", "restart", f"{self.LOGSTASH_CONTAINER_NAME}"])
+        print(f"Logstash container restart result: {result}")
+        if result.returncode != 0:
+            raise Exception(f"Error occurred while reloading Logstash container, see logs for details.")
+
+    def __update_pipeline_config(self):
```
mashhurs, line 126: review note: this requires Logstash run with `--config.reload.automatic` and `elastic-package` config is read-only now, will change.

.buildkite/scripts/run_e2e_tests.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+set -euo pipefail
+
+VERSION_URL="https://storage.googleapis.com/artifacts-api/releases/current"
```
mashhurs, line 5: @dliappis 🎉 initiative!

mashhurs: I have requested a review, I would appreciate if you can review at least the pipeline setup and shell scripts. Python scripts are big but based on KISS since alot logics. As much possible, tried to add comments for edge cases, others read source as going.
Thanks.

.buildkite/e2e-pipeline.yml
```diff
@@ -0,0 +1,14 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+
+agents:
+  provider: "gcp"
+  machineType: "n1-standard-4"
```
dliappis, line 5: Let's use `n2-standard-4` because the `n1` family is quite old and gradually having out of capacity issues

.buildkite/e2e-pipeline.yml
```diff
@@ -0,0 +1,14 @@
...
+  image: family/core-ubuntu-2204
```
dliappis, line 6: Should we use one of the Logstash images instead e.g. `family/platform-ingest-logstash-ubuntu-2204` so that we can benefit from preinstalled things like `ruby`, `java` and tools like `retry` etc?

catalog-info.yaml
```diff
@@ -125,3 +125,50 @@ spec:
...
+          message: "Run plugin E2E tests weekly."
+      teams:
+        logstash:
+          access_level: MANAGE_BUILD_AND_READ
+        ingest-fp:
+          access_level: MANAGE_BUILD_AND_READ
+        ingest-eng-prod:
+          access_level: MANAGE_BUILD_AND_READ
+        everyone:
+          access_level: READ_ONLY
```
dliappis, line 174: I suggest that we commit the changes to `catalog-info.yaml` in a separate PR (pointing to a dummy pipeline file that just emits `hello world` or something); once it gets merged, we can trigger builds from this PR to get validation about the work here.

dliappis (commented):
@mashhurs this is really high quality work, thank you!

I didn't try running it locally yet, but did a first sweep and left some comments.

catalog-info.yaml
mashhurs: Separated in https://github.com/elastic/logstash-filter-elastic_integration/pull/125. Since I don't have concerns about pipeline definition (which is similar to we have with other pipelines), I have included Plugin build scripts to separate from this BIG change to focus only E2E tests.

.buildkite/e2e-pipeline.yml
mashhurs: Definitely, this is I wanted to clarify with you. I saw we have two GCP & AWS similar images. I have updated in a separated [PR](https://github.com/elastic/logstash-filter-elastic_integration/pull/125)

.buildkite/e2e-pipeline.yml
mashhurs: Addressed in a separate PR: https://github.com/elastic/logstash-filter-elastic_integration/pull/125

.buildkite/scripts/e2e/bootstrap.py
mashhurs: The relative: https://github.com/elastic/elastic-package/pull/1668

.buildkite/scripts/run_e2e_tests.sh
```diff
@@ -0,0 +1,40 @@
...
+  echo "Resolved version: $version"
+  export STACK_VERSION="$version"
+}
+
+resolve_current_stack_version
+
+###
+# Build the plugin, to do so we need Logstash source
+build_logstash() {
+  git clone https://github.com/elastic/logstash.git
```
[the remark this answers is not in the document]
mashhurs: Addressed in a separate PR: https://github.com/elastic/logstash-filter-elastic_integration/pull/125

.buildkite/scripts/e2e/bootstrap.py
```diff
@@ -0,0 +1,177 @@
...
+    def __download_elastic_package(self):
+        response = Util.call_url_with_retry(self.ELASTIC_PACKAGE_DISTRO_URL)
+        release_info = response.json()
+
+        download_urls = [asset["browser_download_url"] for asset in release_info["assets"]]
+        download_url = [url for url in download_urls if self.distro in url][0]
+        if download_url.strip() == '':
+            raise Exception("Could not resolve elastic-package distro download URL.")
+
+        os.system(f"wget {download_url}")
```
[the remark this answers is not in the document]
mashhurs: The interesting thing here, I tried downloading with `urllib3` and file downloaded file size differs.
Downloaded size with `urllib3`: 73,661,780 bytes (84.8 MB on disk)
Actual file size: `36,830,890 bytes (36.8 MB on disk)`

It is same with the script you provided as well. In the logs I can see it downloads same size with origin:
```
Download progress: 100.00% (36830890/36830890 bytes)
```
couldn't find the reason so far, with `wget` it is alright.
For now moving to `subprocess`...

.buildkite/scripts/e2e/bootstrap.py
```diff
@@ -0,0 +1,177 @@
...
+        print("elastic-package is successfully downloaded.")
+
+        # Extract the downloaded tar.gz file
+        with tarfile.open(download_url.split("/")[-1], "r:gz") as tar:
+            tar.extractall()
```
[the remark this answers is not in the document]
mashhurs: TIL: didn't know it is slow, moving to `subprocess`.

.buildkite/scripts/e2e/bootstrap.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def __make_elastic_package_global(self):
+        result = subprocess.run(["sudo", "mv", "elastic-package", "/usr/local/bin"])
```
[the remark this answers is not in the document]
mashhurs: Correct me if I am wrong: my 5-6 yrs old memory says non using shell=True is the best practise because it will be OS dependant and shell executions injection processes may fail regardless of current task.
I couldn't find proper blog/guide what changed recently and it is still same. ChatGPT says it is still not recommended 😆😁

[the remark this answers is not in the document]
dliappis: Good questions.
Re: os dependency, do you we plan to run this on >1 different OSes?
The reason I suggested to use `shell=True`, at times, is a) in case you want to bundle things like pipes into your commands b) rely on preexisting env vars.

If you have tested that everything works fine without `shell=True`, e.g. starting up `elastic-package` feel free to ignore the comment and leave it as is.

.buildkite/scripts/e2e/bootstrap.py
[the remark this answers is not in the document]
dliappis: Oh interesting! Obviously there are some bug in this quickly hacked up piece of code. `subprocess` is fine, please us `curl` including retries, if you can. (e.g. `curl --retry 5 --retry-delay 5 -O -fsSL theurl`)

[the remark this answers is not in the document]
mashhurs: Multiple tests (with weak internet connection) shows `subprocess` with `curl --retry` is the better practise for this case, updated.

.buildkite/scripts/e2e/bootstrap.py
```diff
@@ -0,0 +1,177 @@
...
+
+        plugin_name = f"logstash-filter-elastic_integration-{version.strip()}-java.gem"
+        container = Util.get_logstash_container()
+
+        tar_data = io.BytesIO()
+        with tarfile.open(fileobj=tar_data, mode='w') as tar:
+            tar.add(plugin_name, arcname=f"/usr/share/logstash/{plugin_name}")
+
+        tar_data.seek(0)
+        container.put_archive('/', tar_data.getvalue())
```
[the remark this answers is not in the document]
mashhurs: Didn't get much performance difference but from maintenance point of view, moved to subprocess.

.buildkite/scripts/e2e/bootstrap.py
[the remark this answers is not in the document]
mashhurs: I have once more went through `subprocess` package and do see it is pretty well abstracted. However, with _os-dependant_ `Shell=True` I may break something especially when testing on local (mac) and then push to our BK agents. If it is _strong_ suggestion I am open to change and I do believe you have stronger experience on these kinda cases than me.

.buildkite/scripts/e2e/bootstrap.py
```diff
@@ -0,0 +1,150 @@
...
+        # Extract the downloaded tar.gz file
+        Util.run_or_raise_error(["tar", "zxf", file_name],
+                                f"Error occurred while unzipping {file_name}")
+
+    def __make_elastic_package_global(self) -> None:
+        Util.run_or_raise_error(["sudo", "mv", "elastic-package", "/usr/local/bin"],
+                                "Could not make `elastic-package` global")
+
+    def __clone_integrations_repo(self) -> None:
+        Util.run_or_raise_error(["retry", "-t", "5", "--", "git", "clone", "--single-branch",
```
mashhurs, line 66: Since I can't verify on my local, I wonder if there is any issue running `retry` on `subprocess`, @dliappis

.buildkite/scripts/e2e/bootstrap.py
```diff
@@ -0,0 +1,150 @@
...
+        Util.run_or_raise_error(["curl", "-o", file_name, "--retry", "5", "--retry-delay", "5", "-fSL", download_url],
+                                "Failed to download elastic-package")
+        print("elastic-package is successfully downloaded.")
+
+        # Extract the downloaded tar.gz file
+        Util.run_or_raise_error(["tar", "zxf", file_name],
+                                f"Error occurred while unzipping {file_name}")
+
+    def __make_elastic_package_global(self) -> None:
+        Util.run_or_raise_error(["sudo", "mv", "elastic-package", "/usr/local/bin"],
```
mashhurs, line 62: root operation 😬 hope it works @dliappis

dliappis: if the paths are right it will work no worries

.buildkite/scripts/e2e/bootstrap.py
dliappis: You can try it locally too but installing `retry`; if you are on macOS there is a port: https://formulae.brew.sh/formula/retry

dliappis: By the way this fails if the repo already exists (e.g. when testing it locally a few times):

```
fatal: destination path 'integrations' already exists and is not an empty directory.
retry: git returned 128, backing off for 10 seconds and trying again...
```

might be good to make it more resilient (but it will likely include a complex invocation of git commands and/or using the shell too)

.buildkite/scripts/e2e/bootstrap.py
```diff
@@ -0,0 +1,150 @@
...
+
+        # elastic-package creates a profile under home directory
+        config_example_file = os.path.join(self.__get_profile_path(), "config.yml.example")
+        config_file = os.path.join(self.__get_profile_path(), "config.yml")
+        self.__create_config_file(config_example_file, config_file)
+        Util.run_or_raise_error(["elastic-package", "profiles", "use", "e2e"],
+                                "Error occurred while creating a profile. Check logs for details.")
+
+    def __install_plugin(self) -> None:
+        with open("VERSION", "r") as f:
```
dliappis, line 95: Do we create this file somewhere?

```
    with open("VERSION", "r") as f:
         ^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'VERSION'
```

dliappis (commented):
Codewise looks good to me

I hit some errors trying to get this running locally.

.buildkite/scripts/e2e/bootstrap.py
mashhurs: Hm., it is under the repo: https://github.com/elastic/logstash-filter-elastic_integration/blob/main/VERSION
Probably you are running the script inside the `.buildkite/scripts/e2e` (hidden part of the guide 🙈, will think re-organize/revise the guide).

mashhurs: I have updated the `.buildkite/scripts/e2e/readme.md`.

.buildkite/scripts/e2e/README.md
```diff
@@ -0,0 +1,54 @@
...
+```shell
+git clone --single-branch https://github.com/elastic/logstash.git
+cd logstash && ./gradlew clean bootstrap assemble installDefaultGems && cd ..
+LOGSTASH_PATH=$(pwd)/logstash
+export LOGSTASH_PATH
+```
+
+2.Run the following command:
+```shell
+./gradlew clean vendor localGem
```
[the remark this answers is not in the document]
dliappis: I tried running it from this repo and got

```
./gradlew clean vendor localGem

FAILURE: Build failed with an exception.

* Where:
Build file '/home/dl/source/elastic/logstash-filter-elastic_integration/build.gradle' line: 151

* What went wrong:
A problem occurred evaluating root project 'logstash-filter-elastic_integration'.
> LOGSTASH_PATH's logstash-core/lib/jars does not provide any logstash-core jars: /home/dl/source/elastic/logstash/logstash-core/lib/jars. Expression: jarSource.empty

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

```

However, I did prepare Logstash as mentioned above i.e. ran `./gradlew clean bootstrap assemble installDefaultGems` and defined `export LOGSTASH_PATH=/path/to/logstash`

.buildkite/scripts/e2e/bootstrap.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def __update_pipeline_config(self) -> None:
+        local_config_file_path = ".buildkite/scripts/e2e/config/"
+        config_file = "serverless_pipeline.conf" if self.project_type == "serverless" else "pipeline.conf"
+        local_config_file = local_config_file_path + config_file
+        container_config_file_path = "/usr/share/logstash/pipeline/logstash.conf"
+        # python docker client (internally uses subprocess) requires special TAR header with tar operations
+        Util.run_or_raise_error(["docker", "cp", f"{local_config_file}",
+                                 f"{self.LOGSTASH_CONTAINER_NAME}:{container_config_file_path}"],
+                                "Error occurred while replacing pipeline config, see logs for details")
```
dliappis, line 131: Where can the logs be found?

I encountered:

```
  File "/home/dl/source/elastic/logstash-filter-elastic_integration/.buildkite/scripts/e2e/util.py", line 35, in run_or_raise_error
    raise Exception(f"{full_error_message}")
Exception: Error occurred while replacing pipeline config, see logs for details
```

.buildkite/scripts/e2e/README.md
```diff
@@ -0,0 +1,55 @@
...
+Otherwise, it assumes test are being run on Linux platform.
+
+#### Stack version
+E2E also requires `STACK_VERSION` (ex: "8.12.0") environment variable in order to test against.
+Make sure to export it before running. In the Buildkite pipeline, this var will be resolved and exported. 
+
+#### Installing dependencies
+Make sure you have python installed on you local
+```bash
+pip install -r requirements.txt
```
jsvd, line 46:
```suggestion
pip install -r .buildkite/scripts/e2e/requirements.txt
```

.buildkite/scripts/e2e/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+In order to run tests with serverless, you also need to export `EC_API_KEY` which is an organization API key to create a project.
+In the pipelines, this will be automatically retrieved from Vault services.
+
+#### Defining a platform
+If you have Mac, you need to let the E2E know with
+```bash
+export E2E_PLATFORM = "macos"
+```
+Otherwise, it assumes test are being run on Linux platform.
```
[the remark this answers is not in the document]
mashhurs: Yup, took this off.

.buildkite/scripts/e2e/README.md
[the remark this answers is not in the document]
mashhurs: Maybe irrelevant talk but a bit clarification here: `STACK_VERSION` decides the version we run Elastic Stacks with `elastic-package`. We do not touch `ELASTICSEARCH_TREEISH` as our expectation is to test _what we have_ (build plugin with HEAD of the `elastic_integration` repo) against the given stack version (the last major release).

.buildkite/scripts/e2e/bootstrap.py
mashhurs: This will be covered by https://github.com/elastic/elastic-package/pull/1668
Long story short, we need changeable pipeline configuration for long term and I am pushing it with this change.
I will update once it gets landed.

.buildkite/scripts/e2e/bootstrap.py
mashhurs: The expectation is to test with _recent_ integrations, I think continuing the test if repo exists would be an ideal here. On the other hand, it is a meaningless retry.

.buildkite/scripts/e2e/bootstrap.py
mashhurs: New version (0.98.0) of `elastic-package` is released and now this is unblocked.

.buildkite/scripts/e2e/bootstrap.py
dliappis: Yes, if you want to implement this, you shouldn't rely on the native `retry` binary, but instead e.g. use a different helper function that re-runs git clone only if stderr doesn't include `fatal: destination path 'integrations' already exists and is not an empty directory.`.

.buildkite/scripts/e2e/README.md
```diff
@@ -0,0 +1,51 @@
...
+#### Installing dependencies
+Make sure you have python installed on you local
+```bash
+pip install -r .buildkite/scripts/e2e/requirements.txt
+```
+
+### Run
+Run the following command from the repo dir:
+```bash
+python3 .buildkite/scripts/e2e/main.py
```
dliappis, line 48: Btw, if I rerun, e.g. after a failure, I get `Error: error creating profile e2e from profile : profile "e2e" already exists
`

which I can work around by `rm -rf ~/.elastic-package`.

Perhaps we should wipe away this profile before reruns? (or make `elastic-package` lenient about it?)

.buildkite/scripts/e2e/util.py
```diff
@@ -0,0 +1,35 @@
+import docker
+import os
+import requests
+import subprocess
+from docker.models.containers import Container
+from requests.adapters import HTTPAdapter, Retry
+
+
+class Util:
```
dliappis, line 9: nit: you are already in a Python module called util, there's no need to add these functions as staticmethod under a class with the same name; they simply could be functions inside this module.

In that case, you'd change https://github.com/elastic/logstash-filter-elastic_integration/pull/122/files#diff-4948beae97f41385b2fa6d42111231d2f691ca2bef117bce62609a7cbee65b45R11 to `import util` and reference the functions like `util.get_home_path()` etc.

dliappis (commented):
Functionally this looks good. I left some optional comments.

When running the tests locally, using the instructions in the README file, I noticed there were some errors in the output:

```
Error: one or more test cases failed
Internal failure happened with nginx, process return code: 1.
--- Test results for package: nginx - START ---
FAILURE DETAILS:
nginx/access default (variant: v1):
[0] field "message" is undefined


╭─────────┬─────────────┬───────────┬───────────────────────┬────────────────────────────────────────────────────────────────────────────────────────┬───────────────╮
│ PACKAGE │ DATA STREAM │ TEST TYPE │ TEST NAME             │ RESULT                                                                                 │  TIME ELAPSED │
├─────────┼─────────────┼───────────┼───────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┼───────────────┤
│ nginx   │ access      │ system    │ default (variant: v1) │ FAIL: one or more errors found in documents stored in logs-nginx.access-ep data stream │ 48.090519211s │
│ nginx   │ error       │ system    │ default (variant: v1) │ PASS                                                                                   │ 36.341966511s │
│ nginx   │ stubstatus  │ system    │ default (variant: v1) │ PASS                                                                                   │ 31.301999294s │
╰─────────┴─────────────┴───────────┴───────────────────────┴────────────────────────────────────────────────────────────────────────────────────────┴───────────────╯
--- Test results for package: nginx - END   ---
Done
```

however, the script finished successfully:

```
...
Stopping elastic-package stack...
Take down the Elastic stack
Done
$ echo $?
0
```

Shouldn't the above test failure have caused the script to fail too?

.buildkite/scripts/e2e/util.py
mashhurs: Makes sense as it contains all static methods now, removed the class and applied the suggestion, thanks.

.buildkite/scripts/e2e/README.md
mashhurs: Applied removing the profile with `elastic-package` itself.

mashhurs:
> Functionally this looks good. I left some optional comments.
> 
> When running the tests locally, using the instructions in the README file, I noticed there were some errors in the output:
> 
> ```
> Error: one or more test cases failed
> Internal failure happened with nginx, process return code: 1.
> --- Test results for package: nginx - START ---
> FAILURE DETAILS:
> nginx/access default (variant: v1):
> [0] field "message" is undefined
> 
> 
> ╭─────────┬─────────────┬───────────┬───────────────────────┬────────────────────────────────────────────────────────────────────────────────────────┬───────────────╮
> │ PACKAGE │ DATA STREAM │ TEST TYPE │ TEST NAME             │ RESULT                                                                                 │  TIME ELAPSED │
> ├─────────┼─────────────┼───────────┼───────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┼───────────────┤
> │ nginx   │ access      │ system    │ default (variant: v1) │ FAIL: one or more errors found in documents stored in logs-nginx.access-ep data stream │ 48.090519211s │
> │ nginx   │ error       │ system    │ default (variant: v1) │ PASS                                                                                   │ 36.341966511s │
> │ nginx   │ stubstatus  │ system    │ default (variant: v1) │ PASS                                                                                   │ 31.301999294s │
> ╰─────────┴─────────────┴───────────┴───────────────────────┴────────────────────────────────────────────────────────────────────────────────────────┴───────────────╯
> --- Test results for package: nginx - END   ---
> Done
> ```
> 
> however, the script finished successfully:
> 
> ```
> ...
> Stopping elastic-package stack...
> Take down the Elastic stack
> Done
> $ echo $?
> 0
> ```
> 
> Shouldn't the above test failure have caused the script to fail too?

This result is coming from internal `elastic-package`, not from `elastic_integration` or `es-output` plugins. If you take a look more logs, you will be able to see how many events _plugins_ processed. Some of events may get rejected by ES and that is caught by `elastic-package`. Every integration behavior is different and _may_ require different setting especially in `es-output` side (ECS compatibility, ILM, datastream, etc...).
So, with current E2E tests, we are _more_ focusing on running integrations with `elastic_integration` plugin, may extend to output plugins. Current failure looks because ECS is enabled by default and the data is in `event.original`.
If any failure happens with this, tests fail, otherwise prints out every possible failures (`elastic-package`, `es-output`, etc...) for better visibility.

.buildkite/scripts/e2e/bootstrap.py
mashhurs: Couldn't come up with a better solution, if I use git directly losing retry or if I use retry. I have changed retry to 3 to find a baseline for now and if it causes frequent issue, checking existence of `integrations` folder and removing seems a better one for me.

.buildkite/scripts/run_e2e_tests.sh
```diff
@@ -44,8 +44,9 @@ build_logstash
...
 ###
-# TODO: disable ls command and enable with Python E2E scripts
-ls -a
-# Install prerequisites and run E2E tests
-#pip install -r .buildkite/scripts/e2e/requirements.txt
-#python3 .buildkite/scripts/e2e/main.py
\ No newline at end of file
+# Install pip and E2E prerequisites, then run E2E tests
+sudo apt install -y python3-pip
+pip --version
```
mashhurs, line 49: @dliappis it seems `pip` is not available in the VM. I am trying to install it here but facing a prompt blocker. I wonder if we can make built-in pip or overcome prompt. 
image

dliappis: The right command would be `pip3` not `pip`. However, if I were you I'd completely stop using the Python shipped by the Ubuntu 22.04 distro (and thus not `apt install python3-pip` either), but instead leverage one of the pre-installed Python 3 versions via pyenv using a virtualenv, which is the standard way to obtain isolation in the Python ecosystem.

Practically, in your script you'd simply need to specify which version you want to use (e.g. using `pyenv local 3.10.13` which creates a `.python-version` file with this version) and then create a virtualenv using `python3 -mvenv .venv` and activate this using `source .venv/bin/activate`. Note that a standard practice is to commit the `.python-version` in your repo, similar to how we commit `.ruby-version` in Logstash related repos.

With the above steps done, you can use pip with `python3 -mpip ...`. Example below.

```
buildkite-agent@dl-logstash-ubuntu-2204:~$ pyenv versions
* system (set by /opt/buildkite-agent/.pyenv/version)
  3.6.13
  3.10.13
  3.11.5
buildkite-agent@dl-logstash-ubuntu-2204:~$ pyenv local 3.10.13
buildkite-agent@dl-logstash-ubuntu-2204:~$ cat .python-version 
3.10.13
buildkite-agent@dl-logstash-ubuntu-2204:~$ pyenv version
3.10.13 (set by /opt/buildkite-agent/.python-version)
buildkite-agent@dl-logstash-ubuntu-2204:~$ python3 -mvenv .venv
buildkite-agent@dl-logstash-ubuntu-2204:~$ source .venv/bin/activate
(.venv) buildkite-agent@dl-logstash-ubuntu-2204:~$ python3 -mpip install urllib3
Collecting urllib3
  Downloading urllib3-2.2.1-py3-none-any.whl (121 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 121.1/121.1 kB 4.3 MB/s eta 0:00:00
Installing collected packages: urllib3
Successfully installed urllib3-2.2.1

[notice] A new release of pip is available: 23.0.1 -> 24.0
[notice] To update, run: pip install --upgrade pip
```

.buildkite/scripts/e2e/bootstrap.py
mashhurs: Yup, added to the Readme!

dliappis (approved):
LGTM

mashhurs:
🎉 First 🟢 [actual CI](https://buildkite.com/elastic/logstash-filter-elastic-integration-e2e/builds/9#018dcc6c-671f-4808-8fc0-e92abf1a188b)

PR #124, opened 2024-02-02
Fix pipeline slug.

mashhurs:

### DIFF
--- a/.buildkite/pull-requests.json
+++ b/.buildkite/pull-requests.json
@@ -2,11 +2,10 @@
   "jobs": [
     {
       "enabled": true,
-      "pipelineSlug": "logstash-filter-elastic_integration-pull-request-pipeline",
+      "pipelineSlug": "logstash-filter-elastic_integration-pull-request",
       "allow_org_users": true,
       "allowed_repo_permissions": ["admin", "write"],
       "set_commit_status": true,
-      "commit_status_context": "buildkite/logstash-filter-elastic_integration-pull-request-pipeline",
       "build_on_commit": true,
       "trigger_comment_regex": "^(?:(?:buildkite\\W+)?(?:build|test)\\W+(?:this|it))",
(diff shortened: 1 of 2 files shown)


.buildkite/pull-requests.json
```diff
@@ -2,11 +2,10 @@
   "jobs": [
     {
       "enabled": true,
-      "pipelineSlug": "logstash-filter-elastic_integration-pull-request-pipeline",
+      "pipelineSlug": "logstash-filter-elastic_integration-pull-request",
```
dliappis, line 5: to avoid future confusion, we could also rename https://github.com/elastic/logstash-filter-elastic_integration/blob/a71a0700168724e089dd247f01ceb799f082af66/catalog-info.yaml#L39 and https://github.com/elastic/logstash-filter-elastic_integration/blob/a71a0700168724e089dd247f01ceb799f082af66/catalog-info.yaml#L43 to point to the same string: `logstash-filter-elastic_integration-pull-request-pipeline`

.buildkite/pull-requests.json
```diff
@@ -2,11 +2,10 @@
...
       "allow_org_users": true,
       "allowed_repo_permissions": ["admin", "write"],
       "set_commit_status": true,
-      "commit_status_context": "buildkite/logstash-filter-elastic_integration-pull-request-pipeline",
```
dliappis, line 9: this will be the message issued on github; IIRC by default it will be `buildkite/<the pipeline name>` but you can give it some other name.

dliappis (commented):
This should fix the problem, but I left a suggestion to simplify things further i.e. to give the same name for the backstage resource and the pipeline itself.

.buildkite/pull-requests.json
mashhurs: Yup, thank you! I have updated them as well.

PR #125, opened 2024-02-08
E2E pipeline setup with a plugin build shell script.

mashhurs:
### Description

This change setups a BK E2E pipeline for `elastic_integration` plugin. It also includes a shell script to build the plugin before E2E tests start (coming soon with https://github.com/elastic/logstash-filter-elastic_integration/pull/122)

### DIFF
--- a/.buildkite/scripts/run_e2e_tests.sh
+++ b/.buildkite/scripts/run_e2e_tests.sh
@@ -21,7 +21,12 @@ resolve_current_stack_version
 ###
 # Build the plugin, to do so we need Logstash source
 build_logstash() {
-  retry -t 5 -- git clone https://github.com/elastic/logstash.git
+  if [[ -d /usr/local/git-references/git-github-com-elastic-logstash-git ]]; then
+      retry -t 5 -- git clone -v --reference /usr/local/git-references/git-github-com-elastic-logstash-git -- https://github.com/elastic/logstash.git .
+  else
+      retry -t 5 -- git clone --single-branch https://github.com/elastic/logstash.git
+  fi
+
   cd logstash && ./gradlew clean bootstrap assemble installDefaultGems && cd ..
   LOGSTASH_PATH=$(pwd)/logstash
   export LOGSTASH_PATH
--- a/.buildkite/scripts/run_e2e_tests.sh
+++ b/.buildkite/scripts/run_e2e_tests.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+set -euo pipefail
+
+VERSION_URL="https://storage.googleapis.com/artifacts-api/releases/current"
+
+###
+# Resolve stack version and export
+resolve_current_stack_version() {
+  set +o nounset
+
+  local major_version="${ELASTIC_STACK_VERSION%%.*}"
+  local version=$(curl --retry 5 --retry-delay 5 -fsSL "$VERSION_URL/$major_version")
+
+  echo "Resolved version: $version"
+  export STACK_VERSION="$version"
+}
+
+resolve_current_stack_version
+
+###
+# Build the plugin, to do so we need Logstash source
+build_logstash() {
+  retry -t 5 -- git clone https://github.com/elastic/logstash.git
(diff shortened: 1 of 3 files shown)


mashhurs:
@dliappis our latest `pipelineSlug` change didn't fix the CI run :( Any places I can investigate?

mashhurs:
> @dliappis our latest `pipelineSlug` change didn't fix the CI run :( Any places I can investigate?

This was fixed!

.buildkite/scripts/run_e2e_tests.sh
```diff
@@ -0,0 +1,42 @@
...
+  echo "Resolved version: $version"
+  export STACK_VERSION="$version"
+}
+
+resolve_current_stack_version
+
+###
+# Build the plugin, to do so we need Logstash source
+build_logstash() {
+  retry -t 5 -- git clone https://github.com/elastic/logstash.git
```
dliappis, line 24: Our base images have the logstash repo precloned so that we can use the [reference repo](https://git-scm.com/docs/git-clone) feature to speed up cloning and be nice to GitHub wrt bandwidth usage.

What we'd need to do here is something like:

`retry -t 5 -- git clone -v --reference /usr/local/git-references/git-github-com-elastic-logstash-git -- https://github.com/elastic/logstash.git .`

Bonus points if we check first (to allow running this script locally, during development) i.e.

```
if [[ -d /usr/local/git-references/git-github-com-elastic-logstash-git ]]; then
    retry -t 5 -- git clone -v --reference /usr/local/git-references/git-github-com-elastic-logstash-git -- https://github.com/elastic/logstash.git .
else
    retry -t 5 -- git clone https://github.com/elastic/logstash.git
fi
```

catalog-info.yaml
```diff
@@ -125,3 +125,50 @@ spec:
...
+spec:
+  owner: group:ingest-fp
+  type: buildkite-pipeline
+  system: buildkite
+  implementation:
+    apiVersion: buildkite.elastic.dev/v1
+    kind: Pipeline
+    metadata:
+      description: ':logstash: `elastic_integration` E2E tests. :pipeline:'
+      name: logstash-filter-elastic_integration-e2e
```
dliappis, line 151: I recommend we make this name the same as for the resource i.e. `logstash-filter-elastic-integration-e2e` to avoid confusion with things like https://github.com/elastic/logstash-filter-elastic_integration/pull/126/files

dliappis (commented):
Looks great! Left a suggestion for an improvement.

catalog-info.yaml
mashhurs: Definitely agree with your point, updated.
I don't know the risks if we do on other pipelines but I also want to remove `_` from the pipeline definitions beside repo links.

.buildkite/scripts/run_e2e_tests.sh
mashhurs: Good point! Will `--reference` be useful especially on retries or some agents already cloned (couldn't figure out its benefits for our case)?
I have also added `--single-branch` option which I believe it will also somehow contributes to better performance or avoids unnecessary (sor far) fetchings.

dliappis: > Will --reference be useful especially on retries or some agents already cloned (couldn't figure out its benefits for our case)?

Reference repos are a local bare repository whose content is used instead of copying from the remote repository.
If we don't use them, we are unnecessarily downloading a lot of data which is time consuming, error prone and not very kind to GitHub :)

Since the VM images get rebuilt everything week, the content of the logstash reference will have very little drift from the latest HEAD, therefore help in all above mentioned ways.

catalog-info.yaml
dliappis: Given that the process of slugification converts `_` to `-`, I think we should stick to this convention wherever possible, esp `name:` properties in the catalog-info.

dliappis (approved):
LGTM

dliappis (approved):
LGTM

PR #126, opened 2024-02-08
Fix the pipeline slug name.

mashhurs:
### Description
CI doesn't get kicked of because pipeline slug name is different. This PR changes slug name.


jonahbull (approved):
LGTM

PR #127, opened 2024-02-08
fix: preemptively send basic auth

yaauie:
When Elasticsearch is configured to allow anonymous auth, it does not include a challenge header in response to requests, so we cannot rely on the apache http client's auth handling (which is overkill anyway, since an individual Elasticsearch client is not connected to multiple realms).

Instead, we build the http basic auth header ourselves, and register an interceptor to ensure it is added to all requests, similar to how we handle api key auth headers.


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.1.6
+  - Fixes issue where configured `username`/`password` credentials was not sent to Elasticsearch instances that had anonymous access enabled
```
mashhurs, line 2:
```suggestion
  - Fixes issue where configured `username`/`password` credentials was not sent to Elasticsearch instances that had anonymous access enabled [#127](https://github.com/elastic/logstash-filter-elastic_integration/pull/127)
```

mashhurs (approved):
LGTM!
🔴 CI you don't need to care for now. 
- PR should be tested against current release and build pipeline should take care snapshots
- Or if we move to versions truth of source (we discuss in weekly), this will be resolved.

yaauie:
Buildkite failed because it is trying to build against Elasticsearch 8.13's source, but the 8.13 branch has not yet been cut. When building against Elasticsearch 8.11's source (as done in our gradle tasks for building and releasing the artifact), tests succeed.

Issue #120, opened 2024-01-26
[Bug]: Adapt to change in ThreadPool constructor signature


jsvd:
CI is failing with:

```
454.6 > Task :compileJava
--
  | 454.6 /usr/share/plugins/plugin/src/main/java/co/elastic/logstash/filters/elasticintegration/EventProcessorBuilder.java:240: error: no suitable constructor found for ThreadPool(Settings)
  | 454.6             final ThreadPool threadPool = new ThreadPool(settings);
  | 454.6                                           ^
  | 454.6     constructor ThreadPool.ThreadPool(Settings,MeterRegistry,ExecutorBuilder<?>...) is not applicable
  | 454.6       (actual and formal argument lists differ in length)
  | 454.6     constructor ThreadPool.ThreadPool() is not applicable
  | 454.6       (actual and formal argument lists differ in length)
  | 455.0 1 error
  | 455.0
  | 455.0 > Task :compileJava FAILED
```

Due to https://github.com/elastic/elasticsearch/commit/764269b39516f6257d6f994e2e8e4fc66ee68b2d#diff-2835e151b1e7435ff4760a9263e0ecbc359597a9ff02005fc7fec099d38549e6R197

```diff
-    public ThreadPool(final Settings settings, final ExecutorBuilder<?>... customBuilders) {
+    public ThreadPool(final Settings settings, MeterRegistry meterRegistry, final ExecutorBuilder<?>... customBuilders) {
```

Possible fix is to use a NOOP meter registry:

```
-        ThreadPool threadPool = new ThreadPool(settings);
+        ThreadPool threadPool = new ThreadPool(settings, MeterRegistry.NOOP);
```

PR #129, opened 2024-02-09
boundary: use ES API-stable LSIB#createThreadPool

yaauie:
Introduced on Elasticsearch `main` in elastic/elasticsearch#105163 and backported to `8.12` in elastic/elasticsearch#105163. Migrating to this stable API allows our build to work across the 8.12->8.13 boundary in which we observed a change to the ThreadPool's constructor.

Resolves: elastic/logstash-filter-elastic_integration#120

Closes: #121


mashhurs (approved):
🎉 Thanks for making this change.

mashhurs:
Can we version bump and release new version?

yaauie:
> Can we version bump and release new version?

This is so far an internal-only change so I see no reason to release a new version until we have something that is user-facing.

PR #130, opened 2024-02-20
Set Java home to the path where Logstash build requires.

mashhurs:
### Description
This change sets Java home to the path and it is required when building Logstash.


dliappis (approved):
LGTM

PR #131, opened 2024-02-20
Update JRuby version in ruby env.

mashhurs:
### Description
To respect [Logstash core env](https://github.com/elastic/logstash/blob/main/versions.yml#L16), this change updates JRuby version to 9.4.5.0


dliappis (approved):
LGTM

PR #132, opened 2024-02-28
Make GeoIpDatabaseProvider#build public to accommodate JRuby package-private method visibility

jsvd:
Fix for CI failure at runtime: https://buildkite.com/elastic/logstash-serverless-integration-testing/builds/299#018de5b0-39b0-49f3-a295-d15475281758/1258-1265

likely due to  JRuby 9.4.6.0 change:

* Starting with Java 17, package-private methods will no longer be bound automatically even if they can be made visible. [#8061](https://github.com/jruby/jruby/issues/8061), [#8093](https://github.com/jruby/jruby/pull/8093)

https://github.com/jruby/jruby/pull/8093

PR #133, opened 2024-03-06
fix: jruby 9.4.6.0 package visibility strictness

yaauie:
Similar to #132, fixes visibility of `EventProcessorBuilder#build` to work on JRuby 9.4.6.0 which is more strict about making methods visible from ruby.


jsvd (approved):
LGTM

PR #134, opened 2024-04-01
Sync up with ES main to fix the passing whitelists of Painless.

mashhurs:


mashhurs:
Was covered in ES source.

PR #140, opened 2024-05-03
Resolve Logstash container dynamically in E2E tests.

mashhurs:
### Description
Resolve Logstash container dynamically and use it when installing plugin in E2E tests. Before, hard coded container name used, which wasn't failure tolerant.
Example failure: [link](https://buildkite.com/elastic/logstash-filter-elastic-integration-e2e/builds/35#018f45cd-64e3-4ff8-8bb9-dc12d40de347)

### DIFF
--- a/.buildkite/scripts/e2e/util.py
+++ b/.buildkite/scripts/e2e/util.py
@@ -21,7 +21,11 @@ def call_url_with_retry(url: str, max_retries: int = 5, delay: int = 1) -> reque
 
 def get_logstash_container() -> Container:
     client = docker.from_env()
-    return client.containers.get("elastic-package-stack-e2e-logstash-1")
+    containers = client.containers.list(all=True)  # we need to use all to collect logs if container stops
+    for container in containers:
+        if "logstash-" in container.name:  # using only "logstash" may catch logstash ready one
+            return container
+    raise Exception("Logstash container not found")
 
 
 def run_or_raise_error(commands: list, error_message):
--- a/.buildkite/scripts/e2e/util.py
+++ b/.buildkite/scripts/e2e/util.py
@@ -21,7 +21,8 @@ def call_url_with_retry(url: str, max_retries: int = 5, delay: int = 1) -> reque
 
 def get_logstash_container() -> Container:
     client = docker.from_env()
-    containers = client.containers.list(all=True)  # we need to use all to collect logs if container stops
+    # `all=True` catches stopped ones as well, we need all to collect logs if container stops abnormally
+    containers = client.containers.list(all=True)
     for container in containers:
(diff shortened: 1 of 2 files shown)


yaauie:
@mashhurs now that the 8.14 geoip breaking changes have been resolved, can you either rebase onto `main` or merge `main` into this PR?

mashhurs:
> @mashhurs now that the 8.14 geoip breaking changes have been resolved, can you either rebase onto `main` or merge `main` into this PR?

Awesome, thank you so much!
This PR needs uncomment some part (commented for test purpose) nad remove sleep. Will rebase and update!

mashhurs:
E2E CI build: https://buildkite.com/elastic/logstash-filter-elastic-integration-e2e/builds/52

.buildkite/scripts/e2e/util.py
```diff
@@ -21,7 +21,13 @@ def call_url_with_retry(url: str, max_retries: int = 5, delay: int = 1) -> reque
 
 def get_logstash_container() -> Container:
     client = docker.from_env()
-    return client.containers.get("elastic-package-stack-e2e-logstash-1")
+    # `all=True` catches stopped ones as well, we need all to collect logs if container stops abnormally
+    containers = client.containers.list(all=True)
+    for container in containers:
+        print(f"Container name: {container.name}")
+        if "logstash-" in container.name:  # using only "logstash" may catch logstash ready one
```
jsvd, line 28: Is there a compromise between the specificity of `"elastic-package-stack-e2e-logstash-1"` and the potentially ambiguous `"logstash-"`? Maybe in the future there would be another `logstash-` container and this test could fail?

mashhurs: > Is there a compromise between the specificity of "elastic-package-stack-e2e-logstash-1" and the potentially ambiguous "logstash-"?

By default LS container gets `"elastic-package-stack-e2e-logstash-1"` naming but we didn't cover some cases: I have seen BK agent path attached to the container name and/or LS container didn't run due to internal error. With `client.containers.list(all=True)` stopped containers also will be listed and when accessed error clearly tells cannot access to stopped container. If naming is different we still will be able to resolve with `logstash-` [prefix attached by `elastic-agent`](https://github.com/elastic/elastic-package/blob/main/internal/install/application_configuration.go#L109). 

> Maybe in the future there would be another logstash- container and this test could fail?

So far, a single LS container will be spinned by design and if we will have a multiple instances, we need to clarify the use case first and what would be goal I think.

mashhurs:
This issue didn't happen for a while, means docker container name constant to `elastic-package-stack-e2e-logstash-1` as BK agents not changing the path. Closing this issue and will address back if it happens again.

Issue #164, opened 2024-09-27
[Bug]: Plugin Fails Preflight Check When Connecting to HTTPS with SSL Disabled


DumbBoi:
### Plugin Version

logstash-filter-elastic_integration (0.1.13)

### Logstash Version

8.15.2

### Host Info

Linux f36f76d13a8f 5.15.0-122-generic #132~20.04.1-Ubuntu SMP Fri Aug 30 15:50:07 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

### What happened?

The plugin works when connecting to Elasticsearch using HTTP or with a fully configured HTTPS SSL connection, but when I try to run it with `ssl_verification_mode` set to `none` for a HTTPS configured Elasticsearch, the pre-flight check fails with the following traceback (replacing internal uri with placeholders):

```
[2024-09-26T14:42:54,210][ERROR][co.elastic.logstash.filters.elasticintegration.PreflightCheck][events_pipeline] Exception checking serverless: Host name '<SERVICE NAME>' does not match the certificate subject provided by the peer (CN=<INTERNAL SERVICE NAME>, OU=<ORG NAME>)

[...]

PR #165, opened 2024-10-03
ssl: use NoopHostnameVerifier when ssl_verification_mode is 'none'

yaauie:
Fixes: #164

Configuring the trust strategy to trust all certs isn't enough, as that only allows us to trust that the the claims of the certificate _can_ be trusted, but the ssl context still tries to make sure that the trustable cert matches the hostname we requested.

We need to also use a `NoopHostnameVerifier` to bypass hostname matching.


DumbBoi:
Any idea about when this will be included on main ?

mashhurs (approved):
LGTM!

mashhurs:
CI failures (fails on `main` branch, succeeds on 8.15) not are related to this change. `main` failures will be addressed with https://github.com/elastic/ingest-dev/issues/4000

PR #170, opened 2024-10-17
Apply Elasticsearch GeoIP changes and sync with Elasticsearch 8.16.

mashhurs:
### Description
- Reflects https://github.com/elastic/elasticsearch/pull/114250 changes into the plugin
- Updates the Elasticsearch tree to main where plugin pulls when building

The changes are based on the [spike diff](https://gist.github.com/yaauie/0dc9f50f9fa0042c51453376dc0e72da) @yaauie created. This PR additionally validates and adds extra required source changes:
- `MustacheScriptEngine` constructor now requires `Settings` and common settings (`path.home`, `node.home`, etc..)
- `GeoIpProcessor.Factory` also requires type: I have not much knowledge about GeoIP internals but taking a look at the ES source codes there are type for databases (example `ipinfo`) and processors (example `geoip`). And it seems requiring a factory to create `geoip` processor is valid to me.
- Introduces `IpDatabaseAdapter#closeReader()` method to intentionally close the database reader. The `IpDatabase` is an `AutoCloseable` class which automatically closes the database with try-with-resource (`try (IpDatabaseAdapter ipAdapter = new IpDatabaseAdapter())`) operations. Although we have Java 6 style source (`try {...} catch {}...`), auto close is being called when we safely load DB in `ManagedIpDatabaseHolder#loadDatabase()` -> `try {...} catch {..}`

- Closes: #167

PR #178, opened 2024-10-30
geoip: managed databases need to be closed when updated

yaauie:
This is a follow-up to https://github.com/elastic/logstash-filter-elastic_integration/pull/170#pullrequestreview-2400287935, as applied to 8.x

It will need to be forward-ported to `main`.

### DIFF
--- a/src/main/java/co/elastic/logstash/filters/elasticintegration/geoip/ManagedIpDatabaseHolder.java
+++ b/src/main/java/co/elastic/logstash/filters/elasticintegration/geoip/ManagedIpDatabaseHolder.java
@@ -2,6 +2,7 @@ package co.elastic.logstash.filters.elasticintegration.geoip;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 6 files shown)


src/main/java/co/elastic/logstash/filters/elasticintegration/geoip/ManagedIpDatabaseHolder.java
```diff
@@ -54,12 +62,18 @@ public String info() {
...
+            final IpDatabaseAdapter localPreviousDatabase = this.currentDatabase;
             this.currentDatabase = Optional.ofNullable(newDatabasePath)
                     .map(Paths::get)
                     .map(this::loadDatabase)
                     .orElse(null);
+            return localPreviousDatabase;
         });
+
+        if (previousDatabase != null) {
+            IOUtils.closeWhileHandlingException(previousDatabase::closeReader);
```
mashhurs, line 75: This is good catch, thank you!

mashhurs (approved):
LGTM!

PR #180, opened 2024-10-30
prevent releases from `main`

yaauie:
The publisher workflow requires that the `CHANGELOG.md` leads with an entry matching the version, so this will prevent releases on the `main` branch, as desired until we come up with the publish path for Elastic Stack 9.x


mashhurs (approved):
LGTM!

PR #177, opened 2024-10-24
Use target branch in build CI. Properly name the CI jobs.

mashhurs:
### Description
- Introduces a change to use the 8.x branch in build CI.
- Renames the CI jobs to provide proper names

More deeper details about these compatibility concept here:

- with build CI: make sure we keep the plugin compatible with
   - a) any new 8.x releases
   - b) any changes going to be in upcoming future 8.x/minor releases (LS 8.x-SNAPSHOT vs ES main, we may use ES 8.x-SNAPSHOT but we cover it with a) as we are defining the branch in gradle.properties)
   - c) any changes going to be in upcoming future major releases (LS main vs ES main)
- with E2E: make sure the plugin doesn't have an issue when running integrations
   - a) on across 8.x stack
   - b) on across 8.x-SNAPSHOT stack
   - c) on main (or future major version)
- with PR CI: every change introduced under the PR should be compatible across
   - a) current release, granularity LS 8.x vs (ES 8x and ES 8.x-SNAPSHOT)
   - b) upcoming release LS 8.x-SNAPSHOT vs (ES 8.x-SNAPSHOT and ES main)
   - c) upcoming major LS & ES main

Latest CI runs:
- PR: https://buildkite.com/elastic/logstash-filter-elastic-integration-pull-request/builds/375
- Build: https://buildkite.com/elastic/logstash-filter-elastic-integration-build/builds/400
- E2E: https://buildkite.com/elastic/logstash-filter-elastic-integration-e2e/builds/102

### Author's checklist
- [ ] After merge, (nothing functional change but) let's backport to `8.x`

Issue #182, opened 2024-10-31
Re-arranging BK CI jobs.


mashhurs:
### Description
We recently separated branches into 8.x and main but CIs are using both branches to run. With [this PR](https://github.com/elastic/logstash-filter-elastic_integration/pull/177), I tried to bring all CI conditions to 🟢 as possible as quickly. The ideal Jobs state we would like to have with pull request CI
- when any changes happen to 8.x, run 8.x related CI tests
- similarly when any changes happen on main, run main related CI tests

PR #184, opened 2024-11-15
Separete branches in ci

mashhurs:
Backporting https://github.com/elastic/logstash-filter-elastic_integration/pull/177 to 8.x


mashhurs:
@jsvd merging this #177 _backport_ to 8.x w/o waiting for your approval, please raise any your concerns. 
Taking a look at #182 and fix the CI failures affected by version re-definition change.
Also, for the long term, I think adding action to support backporting in this repo would be good.

PR #185, opened 2024-11-17
Fix the CI failures after version definitions in .ci repo.

mashhurs:
This draft PR's goal is to make sure CI status is 🟢 before making any further changes.
- Reflect versions re-definition change: https://github.com/elastic/logstash/blob/main/ci/logstash_releases.json

### DIFF
--- a/.buildkite/build-pipeline.yml
+++ b/.buildkite/build-pipeline.yml
@@ -12,14 +12,11 @@ steps:
     command:
       - .buildkite/scripts/run_tests.sh
     env:
-      ELASTIC_STACK_VERSION: "8.x"
+      ELASTIC_STACK_VERSION: "8.current"
       SNAPSHOT: false
       INTEGRATION: true
       SECURE_INTEGRATION: true
       TARGET_BRANCH: "8.x"
-      # temporary definition to cover PR-170
-      # TODO: remove once 8.16 released
-      ELASTICSEARCH_TREEISH: 8.16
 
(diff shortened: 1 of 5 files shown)


.buildkite/build-pipeline.yml
```diff
@@ -12,22 +12,19 @@ steps:
...
     env:
-      ELASTIC_STACK_VERSION: "8.x"
+      ELASTIC_STACK_VERSION: "8.current"
       SNAPSHOT: false
       INTEGRATION: true
       SECURE_INTEGRATION: true
       TARGET_BRANCH: "8.x"
-      # temporary definition to cover PR-170
-      # TODO: remove once 8.16 released
-      ELASTICSEARCH_TREEISH: 8.16
```
mashhurs, line 22: review note: we no longer need since 8.16 is released.

.ci/docker-run.sh
```diff
@@ -8,7 +8,7 @@ cd .ci
 export BUILDKIT_PROGRESS=plain
 
 if [ "$INTEGRATION" == "true" ]; then
-    docker-compose up --exit-code-from logstash
+    docker compose up --exit-code-from logstash
 else
-    docker-compose up --exit-code-from logstash logstash
+    docker compose up --exit-code-from logstash logstash
```
mashhurs, line 13: review not: we [changed in upstream .ci](https://github.com/logstash-plugins/.ci/blob/1.x/docker-run.sh#L12), doing same here.

jsvd (approved):

PR #187, opened 2024-11-19
Separate pull request CI jobs based on target branch.

mashhurs:
### Description
Only 8.x branch related CI jobs will be kept in a CI yaml definition and removes main branch jobs.

### Limitation
`8.x` has `8.previous` (`8.15` by Nov. 21, 2024) and `8.current` (`8.16` by Nov. 21, 2024). We cannot support a build against ES `8.previous` because ES 8.16 has a different GeoIP interface. If we need to support it, we should probably separate branching.


.buildkite/build-pipeline.yml
```diff
@@ -1,43 +1 @@
-# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
-
```
mashhurs, line 2: we don't need Build and E2E CI jobs definition in 8.x branch as they align on main branch. I am not against for keeping them but it will be clean and clearly tell then we don't use 8.x definitions.

jsvd (approved):
LGTM

PR #194, opened 2024-12-02
Use common .ci -> 1.x as we are doing with rest of plugins.

mashhurs:
### Description
Align on common `.ci` -> `1.x` branch as a standard (other plugins also use).

### Author's checklist
- [x] Run CI after merging following PR and backporting into `1.x` branch: https://github.com/logstash-plugins/.ci/pull/70/files
- [x] Bakckport (cherry pick) this change to `8.x` branch: https://github.com/elastic/logstash-filter-elastic_integration/pull/195

PR #195, opened 2024-12-03
Use common ci 1x 8x

mashhurs:
Cherry pick (backporting) of https://github.com/elastic/logstash-filter-elastic_integration/pull/194


donoghuc (approved):
Backport is clean.

Issue #183, opened 2024-11-12
Add `x-elastic-product-origin` headers to all Logstash requests to Elasticsearch


jsvd:
The Logstash Elasticsearch elastic integration filter plugin makes a number of HTTP requests.

* Investigate the places where HTTP requests are made to Elasticsearch in this plugin
* Add the `x-elastic-product-origin` header to each of those requests.

robbavey:
Updated to correct the header name to x-elastic-product-origin

donoghuc:
Shipped in [0.1.17](https://github.com/elastic/logstash-filter-elastic_integration/releases/tag/v0.1.17)

PR #197, opened 2024-12-05
Set header indicating product origin

donoghuc:
This commit updates the elasticsearch client to set a header indicating which product component is using the API.

Implementation note: The implementation follows the pattern set with setting the elastic API version header. The use of static factory methods for that make the test setup fairly convoluted. I think (with the help of copilot) i've figured out a way to mock everything, but it feels fairly complex.

Closes https://github.com/elastic/logstash-filter-elastic_integration/issues/183


build.gradle
```diff
@@ -51,6 +51,8 @@ configurations {
 
     geolite2 { canBeConsumed = false }
 
+    mockitoAgent { canBeConsumed = false }
```
donoghuc, line 54: See https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.3 for java 21 compat.

donoghuc: In my testing (see java version below), the warnings I was seeing have all been resolved with the changes in `build.gradle`.  

```
openjdk version "21.0.5" 2024-10-15
OpenJDK Runtime Environment Homebrew (build 21.0.5)
OpenJDK 64-Bit Server VM Homebrew (build 21.0.5, mixed mode, sharing)
```

robbavey: Oh cool, that's less invasive than I was envisioning after reading that article

robbavey: And now I read the rest of the `build.gradle` I see the invasive part 🤦‍♂️

robbavey:
Rather than adding under `##UNRELEASED`, now we have a PR to point to, let's up the `minor` version, create a changelog entry for that new version, so we can publish a new version of the plugin - let's get this out there.

mashhurs:
> Rather than adding under `##UNRELEASED`, now we have a PR to point to, let's up the `minor` version, create a changelog entry for that new version, so we can publish a new version of the plugin - let's get this out there.

Yup, we just discussed it and the path forward would be:
- version up and add a changelog with the version: can you please add @donoghuc?
- merge current PR;
- backport against `8.x`;
- adjust if mockito change requires (8.x with JDK17);
- release plugin from 8.x branch (publisher supports the branch selection);

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ## UNRELEASED
   - Releases from `main` are not desired until we get closer to Stack 9.x, and beginning the changelog with something other than release notes matching the version will prevent the plugin release tooling from allowing publication.
 
+## 0.1.17
+    - Add `x-elastic-product-origin` header to Elasticsearch requests [#197](https://github.com/logstash-plugins/logstash-filter-elastic_integration/pull/197)
```
mashhurs, line 5:
```suggestion
    - Add `x-elastic-product-origin` header to Elasticsearch requests [#197](https://github.com/elastic/logstash-filter-elastic_integration/pull/197)
```

mashhurs (dismissed):
LGTM with pull request URL change!

PR #198, opened 2024-12-06
Set header indicating product origin (#197)

donoghuc:
* Set header indicating product origin

This commit updates the elasticsearch client to set a header indicating which product component is using the API.

Implementation note: The implementation follows the pattern set with setting the elastic API version header. The use of static factory methods for that make the test setup fairly convoluted. I think (with the help of copilot) i've figured out a way to mock everything, but it feels fairly complex.

* Release prep for version 0.1.17


mashhurs (approved):
LGTM with 🟢 CI.

PR #199, opened 2024-12-08
[main branch] Rebuilds pull-request, build and E2E pipelines.

mashhurs:
### Description
Reconstructs pull-request, build and E2E pipeline jobs in a way of reflecting recent version expansion.
The pull request script is same as we have in `8.x`. However, build and E2E differ in a way which
- build/E2E periodically (scheduled) visits every branch and validate
- when PR is merged or PR is manually kicked through BK pipeline, it runs with PR changes against `main` 

Results:
- pull-request pipeline:
  - when PR is raised: https://buildkite.com/elastic/logstash-filter-elastic-integration-pull-request/builds/461
- build pipeline
  - periodically (every day): impossible to test unless this PR landed (can be kicked off `main` branch through UI)
  - manually PR kick off (or PR merge): https://buildkite.com/elastic/logstash-filter-elastic-integration-build/builds/478
- E2E pipeline
  - periodically (weekly): impossible to test unless this PR landed (can be kicked off `main` branch through UI)
  - manually PR kick off (or PR merge): https://buildkite.com/elastic/logstash-filter-elastic-integration-e2e/builds/145


yaauie:
Instead of having the various step-builders iterate over [logstash's 
 `ci/logstash_releases.json@main`](https://github.com/elastic/logstash/blob/main/ci/logstash_releases.json) and programmatically determine which to run, I would prefer to have the canonical source be a branch-specific config file with the _list_ of things to run _if they can be resolved_. This inversion keeps the mapping of what Logstashes are supported by the plugin local to the plugin's source which _cannot_ drift.

For example, on any `8.${x}` branch, we always want to test PR's against the upcoming, current, and previous supported 8.x versions, but we don't _care_ about the plugin's ability to run against an unrelesaed possibly-breaking Logstash 9, so the config file would look like:

> `.buildkite/pull-request-matrix.yaml@8.x`
> ~~~ yaml
> releases:
>  - 8.current
>  - 8.previous
> snapshots:
>  - 8.future
>  - 8.current
>  - 8.previous
> ~~~

And on the `main` branch (9.x) before the 9.0 FF and 8._last_ FF, we need to only test against the snapshots for unreleased Logstash 9 and Logstash 8._future_, but since we tolerate null-derefs, we can preemptively add everything so that it will _just work_ once we start getting `9.x` snapshots and releases:

> `.buildkite/pull-request-matrix.yaml@main` (9.x)
> ~~~ yaml
> releases:
>  - 9.current # null-deref-ok
>  - 9.previous # null-deref-ok
> snapshots:
>  - main
>  - 9.future # null-deref-ok
>  - 9.current # null-deref-ok
>  - 9.previous # null-deref-ok
>  - 8.future
> ~~~

The only really _interesting_ but is that once the 8._last_ branch has been cut and 8._future_ disappears, we will need to remap our `8.future` snapshot reference to `8.current`. 

> `.buildkite/pull-request-matrix.yaml@main` (9.x)
> ~~~ diff
>  releases:
>   - 9.current # null-deref-ok
>   - 9.previous # null-deref-ok
>  snapshots:
>   - main
>   - 9.future # null-deref-ok
>   - 9.current # null-deref-ok
>   - 9.previous # null-deref-ok
> - - 8.future
> + - 8.current
> ~~~

And then once 8.last _has_ a release we will need to add it to the matrix for main:

> `.buildkite/pull-request-matrix.yaml@main` (9.x)
> ~~~ diff
>  releases:
>   - 9.current # null-deref-ok
>   - 9.previous # null-deref-ok
> + - 8.current
>  snapshots:
>   - main
>   - 9.future # null-deref-ok
>   - 9.current # null-deref-ok
>   - 9.previous # null-deref-ok
>   - 8.current
> ~~~

.buildkite/scripts/pull-request-pipeline/generate-steps.py
```diff
@@ -0,0 +1,105 @@
+import os
+import requests
+import sys
+import typing
+from requests.adapters import HTTPAdapter, Retry
+
+from ruamel.yaml import YAML
+
+RELEASES_URL = "https://raw.githubusercontent.com/elastic/logstash/main/ci/logstash_releases.json"
+TEST_MATRIX_URL = "https://gist.githubusercontent.com/mashhurs/c5b66e05eac1c0968f841e489a8b43a6/raw/eb6b19174977aa4fbe6f7309e35481fb31690e88/tes.yaml"
```
mashhurs, line 10: review note: this will be updated once https://github.com/elastic/logstash-filter-elastic_integration/pull/202 is landed. This is a test that actual CI steps will be seen here.

mashhurs, line 10:
```suggestion
TEST_MATRIX_URL = "https://raw.githubusercontent.com/elastic/logstash-filter-elastic_integration/main/.buildkite/pull-request-test-matrix.yml"
```

donoghuc (approved):
Thanks for linking the builds based on these.

mashhurs:
> Thanks for linking the builds based on these.

Yup, thank you!
As stated in the description, we cannot run scheduled builds/E2E-tests (confirmed them on my local but BK env vars cannot be involved). The only manual way is to land this change and kick off CIs on BK ui with `main` branch. I will merge and kick them off, if there is anything strange, I will follow up.

PR #200, opened 2024-12-09
[8.x] Backport of #190, dynamic CI pr/build/e2e jobs for target branch.

mashhurs:
### Description
- Backport of #190 into 8.x
- Fixes the CI failure which was using `8.current`, correct setting is to utilize `8.future`, adding a dynamic script to be same in all branches
- adds build and E2E tests which will be kicked off when PRs get merged. Periodic build and E2E tests will be added separately on main branch.

CI test runs
- PR (itself) CI run: https://buildkite.com/elastic/logstash-filter-elastic-integration-pull-request/builds/449
- Build CI run: https://buildkite.com/elastic/logstash-filter-elastic-integration-build/builds/467
- E2E CI run: https://buildkite.com/elastic/logstash-filter-elastic-integration-e2e/builds/140

### Cherry pick details

* Sync up with ES main that PipelineConfiguration class getConfigAsMap() method renamed to getConfig().
* Request modifiable donfig when creating a pipeline.

PR #202, opened 2024-12-10
A matrix to define compatibility across stack versions, where CI job generator leverages

mashhurs:
### Description
When we raise PR, we need to test the plugin changes against its base branch across multiple stacks to make sure we do not break the plugin. This PR introduces the pull-request matrix where all `8.series` branches will align on `8.x` map and `main` with `main` accordingly. The matrix entry for `main` covers the upgrade path where `8.last.last` (will be maintained through release cycle) will be always tested against next major, `9.x/main` for now.

[Dynamic job generator PR](https://github.com/elastic/logstash-filter-elastic_integration/pull/199) applies this change with `YAML()` tool which easifies out life.
```
    matrix_map = call_url_with_retry(MATRIX_MAP)
    matrix_map_yaml = YAML().load(matrix_map.text)
    print(f"matrix_map: {matrix_map_yaml['8.x']}")
```

### DIFF
--- a/.buildkite/pull-request-test-matrix.yml
+++ b/.buildkite/pull-request-test-matrix.yml
@@ -1,0 +1,22 @@
+# a map defines the compatibility between versions
+# some of them might not exist,
+#  and they do not break CI which jobs generator validates the release/snapshot when generating the CI steps
+8.x:
+  releases:
+    - 8.previous
+    - 8.current
+  snapshots:
+    - 8.previous
+    - 8.current
+    - 8.next
+    - 8.future
+main:
+  releases:
+    - 9.previous
+    - 9.current
+  snapshots:
+    - 9.previous
+    - 9.current
+    - 9.next
+    - 9.future
+    - main
--- a/.buildkite/pull-request-test-matrix.yml
+++ b/.buildkite/pull-request-test-matrix.yml
@@ -15,6 +15,7 @@ main:
     - 9.previous
     - 9.current
   snapshots:
+    - 8.future
     - 9.previous
     - 9.current
     - 9.next


donoghuc (commented):
So we are going to encode all the data in both branches? I was thinking these would be managed by branch. So `main` would not have anything about the 8.x and similarly 8.x would not have anything in main. 

I suppose i could see a world in which updating just main and backporting to copy the whole thing to 8.x may be nice, otherwise it would be two separate PRs.

mashhurs:
> So we are going to encode all the data in both branches? I was thinking these would be managed by branch. So `main` would not have anything about the 8.x and similarly 8.x would not have anything in main.
> 
> I suppose i could see a world in which updating just main and backporting to copy the whole thing to 8.x may be nice, otherwise it would be two separate PRs.

We maintain the matrix _only_ in main branch and static URL of `main` will be used. This will be called out in the release guide, so no need backporting.

donoghuc:
I guess I tend to think of branches containing full information for running its CI. The idea that 8.x will be cloned, then a script from that cloned/checked out repo will reach out to github to get data from another branch is not a pattern i'm familiar with. The idea that branches depend on data from other branches is a bit unsettling to me, but it might just be that i'm unfamiliar with an established pattern...

Unrelated: How does step generation/execution handle versions that dont exist? I ask because for the main section it seems like many of those versions do not exist or are all the same, are we able to identify that early and not do too much work?

mashhurs:
> I guess I tend to think of branches containing full information for running its CI. The idea that 8.x will be cloned, then a script from that cloned/checked out repo will reach out to github to get data from another branch is not a pattern i'm familiar with. The idea that branches depend on data from other branches is a bit unsettling to me, but it might just be that i'm unfamiliar with an established pattern...

It is true somehow confusing 🤦 that should be branch specific but we can think of this is a place to keep a global map. We have similar maps such as [`logstash_release.json`](https://github.com/elastic/logstash/blob/main/ci/logstash_releases.json) and [`branches.json`](https://github.com/elastic/logstash/blob/main/ci/branches.json) where we keep maintain in the main branch and widely used, not a branch specific.

Let's discuss with the team if we can move them to a central repository (create it or use simple S3, etc..). What do you think?

> 
> Unrelated: How does step generation/execution handle versions that dont exist? I ask because for the main section it seems like many of those versions do not exist or are all the same, are we able to identify that early and not do too much work?

I have [a safeguard logic](https://github.com/elastic/logstash-filter-elastic_integration/pull/200/files#diff-93bfdeb076ddd5114087ece62e5d34dfb64de1e47127c9ecbccc5842f18d715fR90-R93) which will become no-op if versions doesn't exist. If you check now, `8.next` should be `null` (or doesn't exist) as we have just released `8.17`.

donoghuc (approved):
Looks good. We can discuss the comments without blocking forward momentum on this!

PR #203, opened 2024-12-12
Add a PR backporter

mashhurs:
### Description

This PR adds a PR backporter which we need to backport changes to designated base branches when we merge PRs to target branches.

### DIFF
--- a/.github/workflows/pr_backporter.yml
+++ b/.github/workflows/pr_backporter.yml
@@ -1,0 +1,49 @@
+name: Backport PR to another branch
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  pull-requests: write
+  contents: write
+
+jobs:
+  pr_commented:
+    name: PR comment
+    if: github.event.issue.pull_request
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions-ecosystem/action-regex-match@v2
+        id: regex-match
+        with:
+          text: ${{ github.event.comment.body }}
+          regex: '^@logstashmachine backport (main|[x0-9\.]+)$'
+      - if: ${{ steps.regex-match.outputs.group1 == '' }}
+        run: exit 1
+      - name: Fetch logstash-core team member list
+        uses: tspascoal/get-user-teams-membership@v1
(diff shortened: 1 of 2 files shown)


mashhurs:
> Is there a reason we can just use an off the shelf solution for backport? https://github.com/tibdex/backport seems to be the most popular.

We have a common script in LS core for this purpose: https://github.com/elastic/logstash/blob/main/devtools/backport
I am open to use https://github.com/tibdex/backport or any other new tool but it may require extra processes/effort. For example, our devtools backlporter is fully automated, creates label and attaches to the PR. As I do see with https://github.com/tibdex/backport, we need to manually create label or automate it by converting current python code into nodejs?! (just checked high level details). `tibdex/backport` seems simplified interactions with the GH API, where we are directly using them in our backporter.

**Do you have any concerns with the LS core devtool/backporter?**

Honestly, I haven't seen any pain points so far with LS core backporter. If we have, I would support migrating all LS core and other plugin backporters (if we have) altogether.

donoghuc (approved):
My only concern  apart from general maintainability issues using our own script vs consuming a popular one is copying around files that need updates in the body https://github.com/elastic/logstash-filter-elastic_integration/pull/203/files#diff-38840c5e92fffb039a64f7632661853ab0075df02cb3f3dc123f1de961936a27R100 

I'm fine with just doing this for now while we evaluate benefits of using a pre built action :)

mashhurs:
> My only concern apart from general maintainability issues using our own script vs consuming a popular one is copying around files that need updates in the body https://github.com/elastic/logstash-filter-elastic_integration/pull/203/files#diff-38840c5e92fffb039a64f7632661853ab0075df02cb3f3dc123f1de961936a27R100
> 
> I'm fine with just doing this for now while we evaluate benefits of using a pre built action :)

Ah from that perspective makes sense. Let me create a placeholder to push forward. For now I am not seeing its urgency or broken features, it is nice to have tech dept!

mashhurs:
@logstashmachine backport 8.x

mashhurs:
@logstashmachine backport 8.x

PR #206, opened 2024-12-13
Generate noop CI step if next (is only available during the FF) is N.A

mashhurs:
### Description
Oops, snapshot next stack version can be null since it is only available during the FF. I covered it in the pull-request pipeline and forgot noop placeholder in build/E2E pipeline where this PR covers.

### CI state
- we don't need to wait or doesn't relate to pull-request CI changes

### DIFF
--- a/.buildkite/scripts/build-pipeline/generate-steps.py
+++ b/.buildkite/scripts/build-pipeline/generate-steps.py
@@ -42,7 +42,7 @@ def generate_steps_for_scheduler(versions) -> list:
     steps: list = []
     snapshots = versions["snapshots"]
     for snapshot_version in snapshots:
-        if snapshots[snapshot_version] is None or snapshots[snapshot_version].startswith("7."):
+        if snapshots.get(snapshot_version,"").startswith("7."):
             continue
         full_stack_version = snapshots[snapshot_version]
         version_parts = snapshots[snapshot_version].split(".")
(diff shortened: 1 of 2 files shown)


.buildkite/scripts/e2e-pipeline/generate-steps.py
```diff
@@ -38,7 +38,7 @@ def generate_steps_for_scheduler(versions) -> list:
     steps: list = []
     snapshots = versions["snapshots"]
     for snapshot_version in snapshots:
-        if snapshots[snapshot_version].startswith("7."):
+        if snapshots[snapshot_version] is None or snapshots[snapshot_version].startswith("7."):
```
donoghuc, line 41:
```suggestion
        if snapshots.get(snapshot_version,"").startswith("7."):
```
Nit, but this looks like the use case `get` is meant to solve.

donoghuc (approved):
I think this is fine, suggest using `get` for this https://github.com/elastic/logstash-filter-elastic_integration/pull/206/files#r1884665016 ,  but dont block on that if you need to update in a ton of places 😅

.buildkite/scripts/e2e-pipeline/generate-steps.py
mashhurs: Ah right, makes sense!

.buildkite/scripts/build-pipeline/generate-steps.py
```diff
@@ -42,7 +42,7 @@ def generate_steps_for_scheduler(versions) -> list:
     steps: list = []
     snapshots = versions["snapshots"]
     for snapshot_version in snapshots:
-        if snapshots[snapshot_version].startswith("7."):
+        if snapshots[snapshot_version] is None or snapshots[snapshot_version].startswith("7."):
```
mashhurs, line 45:
```suggestion
        if snapshots.get(snapshot_version,"").startswith("7."):
```

donoghuc (approved):
LGTM!

PR #217, opened 2025-01-08
Rename GH secret.

mashhurs:

### DIFF
--- a/.github/workflows/pr_backporter.yml
+++ b/.github/workflows/pr_backporter.yml
@@ -27,7 +27,7 @@ jobs:
           username: ${{ github.actor }}
           organization: elastic
           team: logstash
-          GITHUB_TOKEN: ${{ secrets.READ_ORG_SECRET_JSVD }}
+          GITHUB_TOKEN: ${{ secrets.READ_ORG_SECRET_MASHHURS }}
       - name: Is user not a core team member?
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_backporter.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           username: ${{ github.actor }}
           organization: elastic
           team: logstash
-          GITHUB_TOKEN: ${{ secrets.READ_ORG_SECRET_JSVD }}
+          GITHUB_TOKEN: ${{ secrets.READ_ORG_SECRET_MASHHURS }}
```
donoghuc, line 30:
```suggestion
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
I dont understand why we need to tie this to someone's personal token. Why cant we use the one provided by GHA? https://docs.github.com/en/enterprise-server@3.15/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

donoghuc: Another bump for "why dont we use an off the shelf solution" https://github.com/elastic/logstash-filter-elastic_integration/pull/203#pullrequestreview-2500611273

donoghuc: Oh, i see the default one applies only to the repo, not the org.

PR #242, opened 2025-01-22
Add an E2E step to scan unsupported processors.

mashhurs:
### Description
We don't have yet the integrations with `elastic_integration` unsupported processors. This PR adds a script to search processors the plugin cannot run.

E2E run: https://buildkite.com/elastic/logstash-filter-elastic-integration-e2e/builds/195

### DIFF
--- a/.buildkite/scripts/e2e-pipeline/bootstrap.py
+++ b/.buildkite/scripts/e2e-pipeline/bootstrap.py
@@ -5,16 +5,26 @@ E2E bootstrapping with Python script
     - Clone integrations repo and prepare packages
     - When E2E finishes, teardown the stack
 """
+import glob
 import os
 import sys
 import time
 import util
+import ruamel.yaml
 
+YAML = ruamel.yaml.YAML()
 
 class Bootstrap:
     ELASTIC_PACKAGE_DISTRO_URL = "https://api.github.com/repos/elastic/elastic-package/releases/latest"
     LOGSTASH_CONTAINER_NAME = "elastic-package-stack-e2e-logstash-1"
     PLUGIN_NAME = "logstash-filter-elastic_integration"
+
+    SUPPORTED_PROCESSORS: list = [
+        "append", "bytes", "community_id", "convert", "csv", "date", "date_index_name", "dissect", "dot_expander",
+        "drop", "fail", "fingerprint", "foreach", "grok", "gsub", "html_strip", "join", "json", "kv", "lowercase",
+        "network_direction", "pipeline", "registered_domain", "remove", "rename", "reroute", "script", "set",
+        "sort", "split", "trim", "uppercase", "uri_parts", "urldecode", "user_agent", "redact", "geoip"
+    ]
 
     def __init__(self, stack_version: str, project_type: str) -> None:
         f"""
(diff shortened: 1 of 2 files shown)


.buildkite/scripts/e2e-pipeline/bootstrap.py
```diff
@@ -5,17 +5,27 @@
...
+import ruamel.yaml
 
+YAML = ruamel.yaml.YAML()
 
 class Bootstrap:
     ELASTIC_PACKAGE_DISTRO_URL = "https://api.github.com/repos/elastic/elastic-package/releases/latest"
     LOGSTASH_CONTAINER_NAME = "elastic-package-stack-e2e-logstash-1"
     PLUGIN_NAME = "logstash-filter-elastic_integration"
 
+    SUPPORTED_PROCESSORS: list = [
```
yaauie, line 22: it would be nice to codify this in a shared yaml that is consumed by [the EventProcessorBuilder](https://github.com/elastic/logstash-filter-elastic_integration/blob/63aa130ba9c76a97bf317c124b5c1f0a5ef1f50e/src/main/java/co/elastic/logstash/filters/elasticintegration/EventProcessorBuilder.java#L94-L131).

Not a blocker, though. Could you file an issue to unify it?

.buildkite/scripts/e2e-pipeline/bootstrap.py
```diff
@@ -73,6 +83,28 @@ def __clone_integrations_repo(self) -> None:
                                  "https://github.com/elastic/integrations.git"],
                                 "Error occurred while cloning an integrations repo. Check logs for details.")
 
+    def __scan_for_unsupported_processors(self) -> None:
+        curr_dir = os.getcwd()
+        pipeline_definition_file_path = "integrations/packages/**/data_stream/**/elasticsearch/ingest_pipeline/*.yml"
+        files = glob.glob(os.path.join(curr_dir, pipeline_definition_file_path))
+        unsupported_processors: dict = {}   # {type: file}
```
yaauie, line 90: should this be a dictionary of arrays, so that we can get _all_ integrations that use an unsupported processor?

.buildkite/scripts/e2e-pipeline/bootstrap.py
```diff
@@ -154,6 +186,7 @@ def run_elastic_stack(self) -> None:
         self.__download_elastic_package()
         self.__make_elastic_package_global()
         self.__clone_integrations_repo()
+        self.__scan_for_unsupported_processors()
```
yaauie, line 189: 🤔 so if the elastic/integrations begin using an unsupported processor, our build breaks. 

I think we might end up needing a "known incompatibilities" list so that we can mark something as known and get our build back to green once we file an issue to address it.

yaauie (changes requested):
:+1:

The only actual issue I have is that we should keep a dict-of-lists-of-filenames instead of a dict-of-filenames.

Other notes are just future improvement ideas.

.buildkite/scripts/e2e-pipeline/bootstrap.py
mashhurs: This is good idea! Otherwise we need to maintain both places. Yup let me see if I can do it with change. Otherwise I will create an issue and follow up later.

.buildkite/scripts/e2e-pipeline/bootstrap.py
mashhurs: Yup this is the idea for the long term. As soon as we will have unsupported plugin (assuming the plugin really cannot support), we will have a white list to continue the E2E without failing. For now, for better visibility breaking CI (**_since it is important to know which integration will become problematic_**) seems better to me.

mashhurs: I will comment this plan in the raising failure part of the source.

.buildkite/scripts/e2e-pipeline/bootstrap.py
mashhurs: Created an issue: https://github.com/elastic/logstash-filter-elastic_integration/issues/249
We can utilize https://github.com/c2nes/javalang to parse the Java class and compare with the processors found from the file. I will follow up and add separately.

.buildkite/scripts/e2e-pipeline/bootstrap.py
mashhurs: Updated with [`1c46477` commit](https://github.com/elastic/logstash-filter-elastic_integration/pull/242/commits/1c464771ae8f6c0d67255fe005dc92d923dc1735)
```
Exception: Unsupported processors found: {'rename_s': ['/logstash-filter-elastic_integration/integrations/packages/1password/data_stream/audit_events/elasticsearch/ingest_pipeline/default.yml', '/logstash-filter-elastic_integration/integrations/packages/1password/data_stream/signin_attempts/elasticsearch/ingest_pipeline/default.yml'], 'json_s': ['/logstash-filter-elastic_integration/integrations/packages/1password/data_stream/audit_events/elasticsearch/ingest_pipeline/default.yml']}
```

yaauie (approved):

mashhurs:
FYI: no need to backport since E2E runs on `main` branch.

PR #248, opened 2025-01-23
mergify: support backports automation with labels

v1v:
## What does this PR do?

* Enable https://docs.mergify.com/workflow for:
  - Creating backports automatically using GitHub backport labels when the original PR gets merged.
  - Notify when backports are stalled.
  - Notify the backports policy
  - Create PRs even with conflicts.
* Enable copy GitHub labels from the original PR to the backported PRs. `Mergify` does not support this feature out of the box, see https://github.com/elastic/apm-server/pull/15190
* Enable merge conflict detection. See https://github.com/elastic/apm-server/pull/15319

## Why is it important/What is the impact to the user?

Use the same backport mechanism as done in some other projects. While still support for the former approach.

### Tasks

- [x] Wait for Mergify to be installed
- [x] Create GitHub labels

### Follow-ups

- Delete https://github.com/elastic/logstash-filter-elastic_integration/blob/main/.github/workflows/pr_backporter.yml

## Related issues

Relates https://github.com/elastic/logstash/issues/16933

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,18 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - 8.*
+      - 9.*
+
+permissions:
+  contents: read
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: elastic/oblt-actions/pre-commit@v1
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,114 @@
+pull_request_rules:
+  - name: ask to resolve conflict
+    conditions:
+      - conflict
+    actions:
+      comment:
+        message: |
+          This pull request is now in conflicts. Could you fix it @{{author}}? 🙏
+          To fixup this pull request, you can check out it locally. See documentation: https://help.github.com/articles/checking-out-pull-requests-locally/
+          ```
+          git fetch upstream
+          git checkout -b {{head}} upstream/{{head}}
+          git merge upstream/{{base}}
+          git push upstream {{head}}
+          ```
+
+  - name: notify the backport policy
+    conditions:
+      - -label~=^backport
+      - base=main
+    actions:
+      comment:
+        message: |
+          This pull request does not have a backport label. Could you fix it @{{author}}? 🙏
+          To fixup this pull request, you need to add the backport labels for the needed
+          branches, such as:
+          * `backport-8./d` is the label to automatically backport to the `8./d` branch. `/d` is the digit.
+          * `backport-8.x` is the label to automatically backport to the `8.x` branch.
+
+  - name: add backport-8.x for the all the PRs targeting main if no skipped or assigned already
+    conditions:
+      - -label~=^(backport-skip|backport-8.x)$
+      - base=main
+    actions:
+      comment:
+        message: |
+          `backport-8.x` has been added to help with the transition to the new branch `8.x`.
+          If you don't need it please use `backport-skip` label.
+      label:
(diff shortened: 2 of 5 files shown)


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,18 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - 8.*
+      - 9.*
```
mashhurs, line 9: Can we leverage the branches [defined in the repo](https://github.com/elastic/logstash/blob/main/ci/branches.json) for CI purpose?

v1v: > Can we leverage the branches [defined in the repo](https://github.com/elastic/logstash/blob/main/ci/branches.json) for CI purpose?

In GitHub actions, the `branches` section accept static values or wildcards , if you mean being explicit, we can, or if you mean to add `7.17`, I can include it. However, I cannot think of any issues for running the `pre-commit` workflow with any push commit in any of those wildcard branches. Or do you have any other concerns?

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,18 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - 8.*
```
v1v, line 8:
```suggestion
      - 7.17
      - 8.*
```

.github/workflows/pre-commit.yml
v1v: BTW, if you want to avoid maintaining  [branches.json](https://github.com/elastic/logstash/blob/main/ci/branches.json), we have a public bucket to help with those details too, see https://storage.googleapis.com/artifacts-api/snapshots/branches.json

mashhurs: My main concern is maintenance:
- do we need to update pre-defined valid `.mergify.yml` on major & minor releases?.
- if so how other teams are modifying to avoid manual effort?

v1v: > do we need to update pre-defined valid .mergify.yml on major & minor releases?.
if so how other teams are modifying to avoid manual effort?


I understand, and thanks for pointing those concerns! This PR is still draft and gathering those questions can help us to have something workable and fully functional.

We have automated that as part of the FF scripts teams use. For instance:

1) APM Server uses this [process](https://github.com/elastic/apm-server/blob/main/dev_docs/RELEASES.md#day-after-feature-freeze)
2) Beats uses an internal process to do the same.

In those cases, there are PRs with those changes created automatically. See https://github.com/elastic/apm-server/pull/11760 or https://github.com/elastic/beats/pull/41706

That's something we can help with, for instance, I added the support for creating the backport labels for the logstash in [elastic/logstash@`f2ccbce` (#16937)](https://github.com/elastic/logstash/pull/16937/commits/f2ccbce41536f60fc6224c55a67abb1d6879ea02). I'll add the support to modify the `mergify` file in logstash and will review how those bumps work for this project to do the same

.mergify.yml
```diff
@@ -0,0 +1,165 @@
...
+          - "{{ author }}"
+        # NOTE: this list needs to be changed when a new minor branch is created
+        #       or an existing release branch reached EOL.
+        branches:
+          - "9.0"
+          - "8.18"
+          - "8.17"
+          - "8.16"
+          - "8.x"
+          - "7.17"
```
mashhurs, line 162: Plugin isn't supported in 7.x
```suggestion
```

mashhurs (approved):
LGTM with minor comment to remove 7.17.

mashhurs (approved):
LGTM!

PR #268, opened 2025-02-03
Simplify running E2E test script on local.

mashhurs:
Simplifies running E2E test script on local by introducing skip set up and integrations to test options.
When first running E2E tests, setup is required. For the next steps setup can be skipped.

### DIFF
--- a/.buildkite/scripts/e2e-pipeline/main.py
+++ b/.buildkite/scripts/e2e-pipeline/main.py
@@ -2,6 +2,7 @@
  Main entry point of the E2E test suites
 """
 
+import argparse
 import os
 from bootstrap import Bootstrap
 from plugin_test import PluginTest
@@ -11,6 +12,10 @@ INTEGRATION_PACKAGES_TO_TEST = ["apache", "m365_defender", "nginx", "tomcat"]
 
 
 class BootstrapContextManager:
+    def __init__(self, skip_setup=False):
+        # save args as attributes
+        self.skip_setup = skip_setup
+
     def __enter__(self):
         stack_version = os.environ.get("ELASTIC_STACK_VERSION")
         project_type = os.environ.get("E2E_PROJECT_TYPE", "on_prems")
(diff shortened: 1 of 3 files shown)


donoghuc (dismissed):
The intent makes sense and the code looks correct. Have some suggestions, though I have not looked at much python recently so my recollections for what is idiomatic may be off.

.buildkite/scripts/e2e-pipeline/main.py
```diff
@@ -57,4 +63,13 @@ def main():
...
 
 if __name__ == "__main__":
-    main()
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--skip-setup')
+    parser.add_argument('--integrations')
+    args = parser.parse_args()
+
+    skip_setup = args.skip_setup == 'true' if args.skip_setup is not None else False
+    integrations = args.integrations.split(',') if args.integrations is not None else []
```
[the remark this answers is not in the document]
mashhurs: Nice points, thanks.

.buildkite/scripts/e2e-pipeline/main.py
```diff
@@ -57,4 +63,13 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--skip-setup, 'action='store_true')
```
donoghuc, line 67:
```suggestion
    parser.add_argument('--skip-setup, 'action='store_false')
```
Oops, i suggested a bug before. Wrong copy paste from the docs.

donoghuc: Specifically i think `store_true` would make the default `True` when we actually want the default to be `False` which i believe corresponds to the `store_false` action.

mashhurs: Ah my bad! Yup, `store_false` makes sense!

donoghuc (approved):
LGTM 🚢

mashhurs:
CI failure is not due to this change, will be fixed by https://github.com/elastic/logstash-filter-elastic_integration/pull/273

PR #273, opened 2025-03-12
Apply upstream ES GeoIP factory interface changes.

mashhurs:
### Description
Recent upstream [ES `GeoIP.Factory#create` interface change](https://github.com/elastic/elasticsearch/pull/124099) broke the plugin.
This PR applies those changes to bring the build state back to green.


andsel (commented):
Maybe my question is naive, but shouldn't we update also https://github.com/elastic/logstash-filter-elastic_integration/blob/27731010dd28154036279c0a7f6cedd1c6b02f79/src/main/java/co/elastic/logstash/filters/elasticintegration/ingest/SetSecurityUserProcessor.java#L38-L39 to include this new ProjectId param?

mashhurs:
> Maybe my question is naive, but shouldn't we update also
> 
> https://github.com/elastic/logstash-filter-elastic_integration/blob/27731010dd28154036279c0a7f6cedd1c6b02f79/src/main/java/co/elastic/logstash/filters/elasticintegration/ingest/SetSecurityUserProcessor.java#L38-L39
> 
> to include this new ProjectId param?

Interesting... you don't see it in the difference?

Screenshot 2025-03-13 at 11 13 21 AM

andsel (approved):
LGTM

PR #279, opened 2025-03-23
Sets default False value to skip-setup flag.

mashhurs:
Our understanding of `'--skip-setup' action='store_true'` seems incorrect that E2E CIs ([example](https://buildkite.com/elastic/logstash-filter-elastic-integration-e2e/builds/204#0195c00d-2767-4bb1-bb45-a0c41d7c3368/60-169)) running with `--skip-setup=True`.
When I take a look at the `store_true`, it seems it automatically _stores_  the default `False` value: https://hg.python.org/cpython/file/2.7/Lib/argparse.py#l861

Similar discussion can be seen here: https://stackoverflow.com/a/8203679

### DIFF
--- a/.buildkite/scripts/e2e-pipeline/main.py
+++ b/.buildkite/scripts/e2e-pipeline/main.py
@@ -64,7 +64,7 @@ def main(skip_setup=False, integrations=[]):
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
-    parser.add_argument('--skip-setup', action='store_false')
+    parser.add_argument('--skip-setup', action='store_true') # store_true adds default False if argument is unavailable.
     parser.add_argument('--integrations')
     args = parser.parse_args()
 


mashhurs:
@v1v I am not able to see the backport-skip label. Do we need to manually create it? If it is up to Mergify, can you please check what happened?
image

v1v:
Likely it was not created. We have centralised the github labels for some projects, and I think the `backport-skip` was not in that list. That's fine, you can create it manually or wait for my contribution to the `elastic/observability-dev` repository to actually create that label - as you prefer. Thanks for the heads up 🙏

mashhurs:
> Likely it was not created either manually or in the automation. Mergify does not create them.
> 
> For the record, we have centralised the github labels for some projects, and I think the `backport-skip` was not in that list. That's fine, you can create it manually or wait for my contribution to the `elastic/observability-dev` repository to actually create that label - as you prefer. Thanks for the heads up 🙏

Now it is here. Thank you alot!

donoghuc (approved):
Oh, my bad. Yeah the "store_" action indicates what will be stored WHEN the flag is set. So in this case when `--skip-setup` is present, the value is `True`.  When i first read that I thout the `store_` action indicated the default value. Its the opposite 🤦

PR #281, opened 2025-03-26
chore: use only mergify for backports

v1v:
After enabling `mergify` for the backports, let's remove the former approach


mashhurs (approved):
LGTM!

v1v:
@mashhurs , I don't have write accss, so I cannot merge it, do you mind merging this PR?

PR #284, opened 2025-04-15
Update VERSION after 8.18.0 stack release

andsel:


jsvd (approved):
LGTM, just be careful when merging not to include the default CLA text blurb

PR #330, opened 2025-07-10
Move to Elasticsearch Stable Bridge API

mashhurs:
## Description

- Create gradle task to compile logstash-bridge module and add jar file to the plugin to make Stable API interfaces available.


build.gradle
```diff
@@ -72,7 +72,7 @@ dependencies {
     elasticsearchClient('co.elastic.clients:elasticsearch-java') {
         // brings latest version available for 8.x branch
         version {
-            strictly('[8.0, 9.0[')
+            strictly('[8.0, 9.0]')
```
mashhurs, line 75:
```suggestion
            strictly('[8.0, 9.0[')
```

build.gradle
```diff
@@ -257,6 +257,45 @@ task buildElasticsearchLocalDistro(dependsOn: unzipDownloadedElasticsearchSource
     }
 }
 
+task buildElasticsearchLogstashBridge(type: Exec) {
+    description "builds logstash-bridge lib module"
+
+    dependsOn buildElasticsearchLocalDistro
+    mustRunAfter buildElasticsearchLocalDistro
```
andsel, line 264: Given that there is already a dependency on `buildElasticsearchLocalDistro` task, this mustRunAfter is redundant I think.

andsel (changes requested):
Left just a comment, if there is real reason to have the `mustRunAfter` please document why.

build.gradle
mashhurs: Yup, I have removed it.

mashhurs:
Plugin build issue is resolved, I have retried the [CI jobs](https://buildkite.com/elastic/logstash-filter-elastic-integration-pull-request/builds/587#_) and they are 🟢 now.

PR #345, opened 2025-07-14
Support terminate processor.

mashhurs:
Registers terminate processor to the support list.

### DIFF
--- a/spec/integration/elastic_integration_spec.rb
+++ b/spec/integration/elastic_integration_spec.rb
@@ -1134,6 +1134,27 @@ describe 'Logstash executes ingest pipeline', :secure_integration => true do
       # end
     end
 
+    describe 'with terminate processor' do
+      let(:pipeline_processor) {
+        '{
+          "terminate": {
+            "if": "ctx.error != null",
+            "tag": "terminated_ingest_pipeline"
+          }
+        }'
+      }
+
+      it 'terminates the ingest pipeline' do
+        events = [LogStash::Event.new(
+          "message" => "Send message to pipeline which gets terminated.",
+          "error" => "This is intentionally placed error.",
+          "data_stream" => data_stream)]
+
+        subject.multi_filter(events).each do |event|
+          expect(event.get("[@metadata][target_ingest_pipeline]")).to include("_none")
+        end
+      end
+    end
   end
 
   context '#multi-pipeline execution' do
--- a/src/main/java/co/elastic/logstash/filters/elasticintegration/EventProcessorBuilder.java
+++ b/src/main/java/co/elastic/logstash/filters/elasticintegration/EventProcessorBuilder.java
@@ -122,6 +122,7 @@ public class EventProcessorBuilder {
                 org.elasticsearch.ingest.common.SetProcessor.TYPE,
                 org.elasticsearch.ingest.common.SortProcessor.TYPE,
                 org.elasticsearch.ingest.common.SplitProcessor.TYPE,
+                "terminate", // note: upstream constant is package-private
                 org.elasticsearch.ingest.common.TrimProcessor.TYPE,
                 org.elasticsearch.ingest.common.URLDecodeProcessor.TYPE,
                 org.elasticsearch.ingest.common.UppercaseProcessor.TYPE,


spec/integration/elastic_integration_spec.rb
```diff
@@ -1134,6 +1134,27 @@ def path; @path; end
       # end
     end
 
+    describe 'with terminate processor' do
+      let(:pipeline_processor) {
+        '{
+          "terminate": {
+            "if": "ctx.error != null",
+            "tag": "terminated_ingest_pipeline"
```
mashhurs, line 1142: I wished this tag be available in `_ingest_document` so that I can `expect(event.get("[tag]")).to eql? "terminated_ingest_pipeline"` but no tag exists.

src/main/java/co/elastic/logstash/filters/elasticintegration/EventProcessorBuilder.java
```diff
@@ -122,6 +122,7 @@ public EventProcessorBuilder() {
                 org.elasticsearch.ingest.common.SetProcessor.TYPE,
                 org.elasticsearch.ingest.common.SortProcessor.TYPE,
                 org.elasticsearch.ingest.common.SplitProcessor.TYPE,
+                "terminate", // note: upstream constant is package-private
```
mashhurs, line 125: `TerminateProcessor` is also package private like `dot_expander` https://github.com/elastic/logstash-filter-elastic_integration/blob/main/src/main/java/co/elastic/logstash/filters/elasticintegration/EventProcessorBuilder.java#L103

spec/integration/elastic_integration_spec.rb
```diff
@@ -1134,6 +1134,27 @@ def path; @path; end
...
+      }
+
+      it 'terminates the ingest pipeline' do
+        events = [LogStash::Event.new(
+          "message" => "Send message to pipeline which gets terminated.",
+          "error" => "This is intentionally placed error.",
+          "data_stream" => data_stream)]
+
+        subject.multi_filter(events).each do |event|
+          expect(event.get("[@metadata][target_ingest_pipeline]")).to include("_none")
```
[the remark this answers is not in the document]
mashhurs: After plugin successfully executes the pipeline, it let's know (set the `[@metadata][target_ingest_pipeline]` to `_none`) [es-output to avoid ingest pipeline execution](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/fc281c3f44718b178a3fdcfe3ddb07c0b16e8edd/lib/logstash/outputs/elasticsearch.rb#L583) in Elasticsearch side.

> what exactly this assertion proves WRT the termination?

Yeah, this one I tried to highlight [here](https://github.com/elastic/logstash-filter-elastic_integration/pull/345#discussion_r2206012415) that `terminate` processor doesn't attach any specific fields we can validate (it seems `tag` is for debug purpose). I have updated the integration test which adds `append` processor after `terminate` and validates to make sure append processor didn't execute.
Correspond ES query as an integration test would like this:
```
POST /_ingest/pipeline/_simulate
{
  "pipeline" :
  {
    "processors": [
      {
        "terminate": {
          "if": "ctx.error != null",
          "tag": "terminated_pipeline"
        }
      },
      {
        "append": {
          "field": "message",
          "value": "my message value"
        }
      }
    ]
  },
  "docs": [
    {
      "_index": "index",
      "_id": "id1",
      "_source": {
        "foo": "bar",
        "error": "some exception"
      }
    },
    {
      "_index": "index",
      "_id": "id2",
      "_source": {
        "foo": "rab"
      }
    }
  ]
}
```

In the result, we can see `id1` will not have `message` since pipeline was terminated by terminate processor:
```
{
  "docs": [
    {
      "doc": {
        "_index": "index",
        "_version": "-3",
        "_id": "id1",
        "_source": {
          "error": "some exception",
          "foo": "bar"
        },
        "_ingest": {
          "timestamp": "2025-07-15T23:56:43.427866851Z"
        }
      }
    },
    {
      "doc": {
        "_index": "index",
        "_version": "-3",
        "_id": "id2",
        "_source": {
          "message": [
            "my message value"
          ],
          "foo": "rab"
        },
        "_ingest": {
          "timestamp": "2025-07-15T23:56:43.427882279Z"
        }
      }
    }
  ]
}
```

donoghuc (approved):
Thanks for the detailed breakdown. That is very helpful. Appreciate the extra context.

PR #376, opened 2025-11-04
Remove no need 8.x versions from CIs tests.

mashhurs:
### Description
Since we no longer have 8.prev releases, this PR removes it from CIs.

Note that all CIs use version definitions from `main` branch, no need to backport.

### DIFF
--- a/.buildkite/scripts/e2e-pipeline/generate-steps.py
+++ b/.buildkite/scripts/e2e-pipeline/generate-steps.py
@@ -38,7 +38,7 @@ def generate_steps_for_scheduler(versions) -> list:
     steps: list = []
     snapshots = versions["snapshots"]
     for snapshot_version in snapshots:
-        if snapshots[snapshot_version] is None or snapshots[snapshot_version].startswith("7.") or snapshot_version == '8.previous':
+        if snapshots[snapshot_version] is None or snapshots[snapshot_version].startswith("7.") or snapshot_version == "8.previous":
             continue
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/e2e-pipeline/generate-steps.py
```diff
@@ -38,7 +38,7 @@ def generate_steps_for_scheduler(versions) -> list:
     steps: list = []
     snapshots = versions["snapshots"]
     for snapshot_version in snapshots:
-        if snapshots[snapshot_version] is None or snapshots[snapshot_version].startswith("7."):
+        if snapshots[snapshot_version] is None or snapshots[snapshot_version].startswith("7.") or snapshot_version == "8.previous":
```
donoghuc: I'm not following this case. What is this for?

mashhurs: This is for E2E pipeline and this filters out 8.previous from `logstash-versions.yml` snapshots entry:
```
# Track the ongoing SNAPSHOT for the "next" release for each stream
snapshots:
  7.current: "7.17.30-SNAPSHOT"
  8.previous: "8.18.9-SNAPSHOT"
  8.current: "8.19.7-SNAPSHOT"
  9.previous: "9.1.7-SNAPSHOT"
  9.current: "9.2.1-SNAPSHOT"
  main: "9.3.0-SNAPSHOT"
```

PR #377, opened 2025-11-20
Play around with E2E tests to improve.

mashhurs:
### Description
This change suggestions:
- Better visibility for the entire stack by printing docker container [LS, ES, elastic-agent] logs
- it was found that with 9.3.0-SNAPSHOT stack version, elastic-agent doesn't use SSL params. Introduced a separate config for such cases
- increases resource/VM-spec to make sure all stack can spin up
- fixes flaky unit test

Test CI: https://buildkite.com/elastic/logstash-filter-elastic-integration-e2e/builds/250#019aa917-7cd2-42bd-9fe9-7a44e840ae53


.buildkite/scripts/e2e-pipeline/bootstrap.py
```diff
@@ -197,8 +197,8 @@ def run_elastic_stack(self, skip_setup=False) -> None:
             self.__scan_for_unsupported_processors()
             self.__setup_elastic_package_profile()
         self.__spin_stack()
-        self.__install_plugin()
-        self.__reload_container()
+        #self.__install_plugin()
+        #self.__reload_container()
```
mashhurs: self-reminder: need to rollback

src/test/java/co/elastic/logstash/filters/elasticintegration/ElasticsearchRestClientWireMockTest.java
```diff
@@ -175,7 +175,10 @@ void setupMock() {
 
             try (RestClient restClient = rcb.build()) {
                 final SSLHandshakeException ex = assertThrows(SSLHandshakeException.class, () -> restClient.performRequest(new Request("GET", "/")));
-                assertThat(ex.getMessage(), stringContainsInOrder("fatal", "bad_certificate"));
+                assertThat(ex.getMessage(), allOf(
+                    containsString("fatal"),
+                    anyOf(containsString("bad_certificate"), containsString("certificate_required"))
```
mashhurs: In some hosts (mostly with recent builds), this test fails with the below error, looks like a JVM specific.
I can't reproduce it on my local (my JVM always uses build in certificates) but taking a look at [RFC-8446](https://datatracker.ietf.org/doc/html/rfc8446), the unit test is not sending any certificates, so `certificate_required` looks a correct validation here.
image

- The CI failure link: https://buildkite.com/elastic/logstash-filter-elastic-integration-pull-request/builds/659#019aa856-f319-4ae0-9c3d-4eba70714a6c/74-1400
- The error log
```
logstash-1       \|   Test testBasicConnectivityDisablingVerification() FAILED
--
logstash-1       \|
logstash-1       \|   java.lang.AssertionError:
logstash-1       \|   Expected: a string containing "fatal", "bad_certificate" in order
logstash-1       \|        but: was "(certificate_required) Received fatal alert: certificate_required"
logstash-1       \|       at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
logstash-1       \|       at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
logstash-1       \|       at co.elastic.logstash.filters.elasticintegration.ElasticsearchRestClientWireMockTest$MutualHttps.testBasicConnectivityDisablingVerification(ElasticsearchRestClientWireMockTest.java:178)
logstash-1       \|       at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
logstash-1       \|       at java.base/java.lang.reflect.Method.invoke(Method.java:580)
```

.buildkite/scripts/e2e-pipeline/util.py
```diff
@@ -18,11 +18,30 @@ def call_url_with_retry(url: str, max_retries: int = 5, delay: int = 1) -> reque
...
+    for container in matching_containers:
+        # pretty printing with clear separators
+        separator = "=" * 80
+        print(f"\n{separator}")
+        print(f"Container: {container.name}")
+        print(f"{separator}")
+        container_logs = container.logs().decode('utf-8')
+        for log_line in container_logs.splitlines():
+            print(f"  {log_line}")
+        print(f"{separator}\n")
```
mashhurs: This gives us better visibility what happened with entire stack.

.buildkite/scripts/e2e-pipeline/bootstrap.py
```diff
@@ -160,9 +161,33 @@ def __reload_container(self) -> None:
...
+            else:
+                return False
+        except ValueError:
+            return False
+
     def __update_pipeline_config(self) -> None:
         local_config_file_path = ".buildkite/scripts/e2e-pipeline/config/"
-        config_file = "serverless_pipeline.conf" if self.project_type == "serverless" else "pipeline.conf"
+        if self.__is_version_gte(self.stack_version, 9, 3):
+            config_file = "pipeline-ea-without-ssl.conf"
```
donoghuc: I cant think of a reason to special case this. I think we should figure out *why* this all the sudden does not use ssl rather than letting it use unencrypted.

donoghuc (changes requested):
I'm good with the added logging, the more flexible assertian for the bad_certificate vs certificate_required, and the more capable VM. But from what i can tell the `elastic_agent` stopping using ssl seems like a legit bug we need to fix rather than encoding the unencrypted communication into our tests. It feel like a legit regression we need to address. Can we split that out?

mashhurs:
> > Triggered a build against the removal of the plaintext for 9.3.0/main commit https://buildkite.com/elastic/logstash-filter-elastic-integration-e2e/builds/253 assuming that is green this is good to go!
> 
> I don't think this will succeed on 9.3-SNAPSHOT since we have issue with elastic-agent policy (not using SSL params when connecting to Logstash).

But I do see the root cause is fixed with https://github.com/elastic/kibana/pull/244624 and it might require elastic-agent, kibana & fleet (elastic-package as well?!) releases. I will follow up on this.

Issue #379, opened 2026-01-14
Unit tests are broken with ES upstream main.


mashhurs:
### Description

When we run unit tests on `main` branch, we get following errors.
It looks like with the [recent change in ES](https://github.com/elastic/elasticsearch/commit/ebc04a6d3f6d8c12120c5cda2fc1addd8bbe4191), `com/carrotsearch/hppc/ObjectObjectHashMap` through the `import org.elasticsearch.index.mapper.Mapping;`.

It looks to me a solution is straightforward to include `hppc` package

```
java.lang.NoClassDefFoundError: com/carrotsearch/hppc/ObjectObjectHashMap
	at org.elasticsearch.common.collect.ImmutableOpenMap.<clinit>(ImmutableOpenMap.java:276)
	at org.elasticsearch.cluster.metadata.Metadata$Builder.<init>(Metadata.java:1360)
	at org.elasticsearch.cluster.metadata.Metadata.builder(Metadata.java:1311)
	at org.elasticsearch.cluster.metadata.Metadata.<clinit>(Metadata.java:187)
	at org.elasticsearch.cluster.metadata.IndexMetadata.<clinit>(IndexMetadata.java:3076)
	at org.elasticsearch.index.IndexSettings.<clinit>(IndexSettings.java:354)
	at org.elasticsearch.index.mapper.Mapper.lambda$static$0(Mapper.java:87)
	at org.elasticsearch.common.settings.Setting.<init>(Setting.java:216)
	at org.elasticsearch.common.settings.Setting.<init>(Setting.java:291)
	at org.elasticsearch.common.settings.Setting.<init>(Setting.java:344)
	at org.elasticsearch.common.settings.Setting.enumSetting(Setting.java:1747)
	at org.elasticsearch.index.mapper.Mapper.<clinit>(Mapper.java:84)
	at org.elasticsearch.index.mapper.RootObjectMapper$Defaults.<clinit>(RootObjectMapper.java:61)
	at org.elasticsearch.index.mapper.RootObjectMapper$Builder.<init>(RootObjectMapper.java:73)
	at org.elasticsearch.index.mapper.Mapping.<clinit>(Mapping.java:36)
	at org.elasticsearch.common.xcontent.XContentHelper.parseToType(XContentHelper.java:225)
	at org.elasticsearch.common.xcontent.XContentHelper.parseToType(XContentHelper.java:212)
[...]

PR #380, opened 2026-01-14
Include required hppc package, fix String.format issue and forwardport branch changelogs

mashhurs:
## Description
While locally running unit tests, the failures show the breakage caused by upstream ES change as described in #379.
In the [build-pipeline](https://buildkite.com/elastic/logstash-filter-elastic-integration-build), we designed to run integration tests which looks like not enough.
This change:
- includes `hppc` required package
- enables unit tests in the build-pipeline, run with this PR - https://buildkite.com/elastic/logstash-filter-elastic-integration-build/builds/921
- Forwardports changelogs after 9.2 and 9.3 plugin releases
- Fixes the bug where `String.format` may raise `MissingFormatArgumentException` in the `SimpleCacheResolver`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## UNRELEASED
+  - Fixes the `MissingFormatArgumentException` potentially `String.format` may cause in the `SimpleResolverCache` [#380](https://github.com/elastic/logstash-filter-elastic_integration/pull/380)
+
+## 9.3.0
+  - Embeds Ingest Node components from Elasticsearch 9.3 [#378](https://github.com/elastic/logstash-filter-elastic_integration/pull/378)
+
+## 9.2.0
+  - Logging compatability with Elasticsearch 9.2 [#373](https://github.com/elastic/logstash-filter-elastic_integration/pull/373)
+  - Utilizes Elasticsearch interfaces via Elasticsearch logstash-bridge [#336](https://github.com/elastic/logstash-filter-elastic_integration/pull/336)
+
 ## 9.1.1
   - Add `terminate` processor support [#345](https://github.com/elastic/logstash-filter-elastic_integration/pull/345)
 


src/main/java/co/elastic/logstash/filters/elasticintegration/resolver/SimpleResolverCache.java
```diff
@@ -98,7 +98,7 @@ public Optional<V> resolve(final K resolveKey,
                     LOGGER.trace(() -> String.format("uncached-load(%s){ %s -> %s }", type, resolveKey, retrieved.getCachedValue()));
                     return (retrieved.isHit() || !retrieved.isExpired()) ? retrieved : null;
                 } catch (Exception e) {
-                    LOGGER.debug(() -> String.format("uncached-load-exception(%s){ %s !> %s }", type, resolveKey), e);
+                    LOGGER.debug(() -> String.format("uncached-load-exception(%s){ %s !> %s }", type, resolveKey, e.getMessage()), e);
```
mashhurs: Not related to this change. I got to know that `String.format` is requiring 3 params but 2-vars were being provided. I don't think this requires a new release as I believe we have upper level catch to avoid plugin crashes.

.buildkite/scripts/build-pipeline/generate-steps.py
```diff
@@ -19,18 +19,22 @@ def call_url_with_retry(url: str, max_retries: int = 5, delay: int = 1) -> reque
     return session.get(url)
 
 
-def generate_test_step(stack_version, branch, snapshot) -> dict:
-    label_integration_test: typing.final = f"Integration test for {stack_version}, snapshot: {snapshot}"
+def generate_test_step(stack_version, branch, snapshot, mark_integration_test = True) -> dict:
```
donoghuc: Two nitpick suggestions:
1. `mark_integration_test` is a confusing name, maybe `is_integration_test` if it is boolean, otherwise passing in a string that is `test_type` that accepts `unit` or `integration`. 
2. I'm not a fan of tacking on trailing optional parameters uness absolutely required as it makes future changes to the method signature harder. Given how low impact this is, i think leaving it as a required parameter is great (it looks like all the consumers update anyway).

donoghuc (changes requested):
Just a couple suggestions https://github.com/elastic/logstash-filter-elastic_integration/pull/380#discussion_r2692231146 

Probably also the bugfix for the string format bug in the debug log message could use a call out in the PR description (maybe even a release note)

donoghuc (approved):
🚀

PR #431, opened 2026-03-14
Autorenew SSL certs and use them everywhere

mashhurs:
Background: Currently, SSL certs are static, means already generated and they have expiration date. They are used in the tests, used scopes are: unit tests, ES docker file and integration tests. 

This PR:
- introduces Autorenew SSL certs and use them everywhere like in java unit tests, docker files, ruby spec and integration tests. One caveat here is that since ES docker file needs to apply SSLs before it starts, plugin needs to create SSL certs for the ES docker file before running integrations tests. 
Places use the certificates are:
  - ES docker file (used by integration tests): `./utils/ssl-certs-generator/generate.sh` shell script generates on flight;
  - specs: `./utils/ssl-certs-generator/generate.sh` shell script generates
  - java unit tests: gradle `generateTestCertificates` generates;
Gradle `clean` and `generate.sh` (before running) cleans up old certificates if exist.

- upgrades junit, wiremock, mokito dependencies used by unit tests 

- Closes #425


.buildkite/scripts/run_tests.sh
```diff
@@ -11,4 +11,6 @@ else
   git checkout "$TARGET_BRANCH"
 fi
 
+./utils/ssl-certs-generator/generate.sh
```
donoghuc, line 14: Given gradle does the cleanup, i think it should also do the generation. Did you consider managing the invocation there? That way it manages certs end to end (creation/deletion).

mashhurs: Oh right, especially for the integration tests (just saw they also failed possibly because of this reason), shell script generated SSLs used by ES and they will be cleaned up by gradle, this is a caveat. Let me think..

mashhurs: I don't like mixing the areas (.ci, gradle) but from the high level context we should generate certs to apply for ES container and do not clean up when building the plugin - https://github.com/elastic/logstash-filter-elastic_integration/pull/431/changes/6ce05cb1cb210e71f8cd5d28b6a1dad81e1527ab

donoghuc: I'm not following... Why is generate.sh invoked *outisde* of gradle. Essentially, i would expect that the orchestration engine for any workflow would be gradle, so expressing the handling of certs should be fully confined to that.

mashhurs: > i would expect that the orchestration engine for any workflow would be gradle

I would love to do so but the situation a bit different here because plugin needs to apply SSL certs in the ES docker file before integrations tests run. The sequence:
- CI host: `run_tests.sh` calls `generate.sh` -> certs created
- Docker build: ES Dockerfile copies those certs, LS Dockerfile copies the whole plugin dir (including certs)
- LS container starts: `bundle exec rake vendor` -> `./gradlew clean vendor` -> certs get deleted by clean
- LS container: gradle's `generateTestCertificates` runs (via `processTestResources`) -> calls `generate.sh` again → regenerates new certs
- ES container: still has the original certs from step 1 -> DANG mismatch...

So, overall I think the best way for current situation is
- to generate certs first in CI host -> `run_tests.sh`: used by both unit and integration tests
- generate in gradle only if needed, e.g. local `./gradlew clean test`
  - we will not clean certs with gradle clean, if we do integration tests will fail due to mismatch
  - and genera.sh will skip if certs exist - makes sure we are not generating twice

WDYT? Anything am I missing or any other ideas to handle better?

donoghuc: I think this is a solid improvment. We can address orchestration cohesion separately. Thanks.

donoghuc (approved):
I think this is a solid improvment, my concerns around gradle consolidation can be addressed seprately. 

Can we note somewhere in the PR description or final squashed commit that dependencies were also bumped as part of this refactor?

mashhurs:
> I think this is a solid improvment, my concerns around gradle consolidation can be addressed seprately.
> 
> Can we note somewhere in the PR description or final squashed commit that dependencies were also bumped as part of this refactor?

Commit included the test dependencies upgrade message and I have more details like which dependencies. I have also added under the PR description as well.

mashhurs:
@mergifyio backport 9.4

PR #445, opened 2026-04-17
[Sync up with ES main] Apply geoip module relocation changes

### LINKED ISSUE #449: Apply upstream ES ingest-geoip module refactoring
There is initiative - https://github.com/elastic/elasticsearch/pull/145011 to refactor geoip-module in a way that any other 3rd party can use it. This change impacts the plugin since it now understands the ingest-geoip as a bundled package/module but with the upstream change the bundle/package will be separated. It depends on what and how changes will be made but most likely plugin needs to build (build might not need if local distro does) required modules and include interfaces in the final jar.

mashhurs:
Pulls this https://github.com/elastic/elasticsearch/pull/145011 PR changes and tests

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -371,7 +371,7 @@ task shadeElasticsearchIngestGeoIpModule(type: com.github.jengelman.gradle.plugi
 
     dependsOn buildElasticsearchLocalDistro
 
-    from(buildElasticsearchLocalDistro.module("ingest-geoip").orElse(objects.fileCollection()))
+    from(buildElasticsearchLocalDistro.module("ip-location").orElse(objects.fileCollection()))
 
     archiveFileName = 'ingest-geoip-shaded.jar'
     destinationDirectory = file("${buildDir}/shaded")
@@ -453,6 +453,7 @@ task importMinimalElasticsearch() {
                 include jarPackageNamed("elasticsearch-x-content")
                 include jarPackageNamed("elasticsearch-geo")
                 include jarPackageNamed("elasticsearch-user-agent-api")
+                include jarPackageNamed("elasticsearch-ip-location-api")
                 include jarPackageNamed("lucene-core")
                 include jarPackageNamed("lucene-analysis-common")
                 include jarPackageNamed("hppc")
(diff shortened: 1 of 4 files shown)


build.gradle
yaauie, line 369:
```suggestion
task shadeElasticsearchIpLocationModule(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar) {
```

build.gradle
```diff
@@ -371,7 +371,7 @@ task shadeElasticsearchIngestGeoIpModule(type: com.github.jengelman.gradle.plugi
 
     dependsOn buildElasticsearchLocalDistro
 
-    from(buildElasticsearchLocalDistro.module("ingest-geoip").orElse(objects.fileCollection()))
+    from(buildElasticsearchLocalDistro.module("ip-location").orElse(objects.fileCollection()))
```
yaauie, line 374: do we need everything, or can we get only the geoip dependencies?

build.gradle
```diff
@@ -472,6 +473,11 @@ task importMinimalElasticsearch() {
             }
 
             from(shadeElasticsearchIngestGeoIpModule)
```
yaauie, line 475:
```suggestion
            from(shadeElasticsearchIpLocationModule)
```

yaauie (commented):
There's some nuance with how we get the shaded dependencies lined up for this project, the new `ingest-ip-location` module, and the _dependencies_ of the `ip-location` module (but maybe not the functionality from that module, since the `ingest-ip-location` provides an implementation of `elasticsearch-ip-location-api` for use with its database providers). But Logstash _also_ provides those dependencies, although it may be a different version which is why we need the shading.

I'll need to validate it next week when my head has the room to think it through.

build.gradle
mashhurs: Yep, properly renamed them now as my first _very minimal_ change was to check the tests.

build.gradle
mashhurs: Yeah, as per my analysis (correct me if I am wrong), we need all (`ip-location`, `ip-location-api` and `ingest-ip-location`) need to exist in the JARs but we have to only shade the `ip-location` since others are references by the logstash-bridge:
- `IpLocationServiceAdapter` from `ingest-ip-location`
- no-op `IpLocationService` from `ip-location-api`

mashhurs:
Just fyi: BK build history is increasing because I am regularly kicking off the BK CI when I notice (subscribed) upstream ES PR change.

donoghuc (approved):
Tests are green and it looks like this vendors and includes the refactored geiop code correctly.

PR #459, opened 2026-05-11
Update changelog and VERSION with latest releases

donoghuc:
This commit updates the VERSION to reflect the next planned minor. It also updates the changelog with release notes for the last two patch releases for 9.4 and 9.3.


VERSION
```diff
@@ -1 +1 @@
-9.4.0
+9.5.0
```
mashhurs, line 1: WDYT if we always keep this UNRELEASED?
My concern is if we run plugin publisher by mistake on main branch, it may still publish plugin with this version since it is not in rubygem.

mashhurs (approved):
LGTM!
Asked your opinion to make one consistent rule (keep unreleased for both version & changelog) but shouldn't be necessarily now.

VERSION
donoghuc: as long as `UNRELEASED` doesn't cause any build failures for testing ETC. If so we could do a nonsense version like 999.999.999 or something.

PR #461, opened 2026-05-14
Support more java timestamp types

### LINKED ISSUE #460: [Bug]: missing converter for `java.util.Date`
### Plugin Version

```shell
all/HEAD
```

### Logstash Version

```shell
all/HEAD
```

### Java Version

```shell
JDK21+
```

### Host Info

```shell
all
```

### What happened?

When an integration sets a field with a value that is an instance of `java.util.Date`, logstash fails to convert the value to a type that is usable on the Event.

Relates: https://github.com/elastic/logstash-filter-elastic_integration/issues/65

yaauie:
- adds support for `java.util.Date` to resolve #460
 - opens support for `java.time.ZonedDateTime` to its superclass `java.time.chrono.ChronoZonedDateTime`
 - adds fast-success support for `java.time.Instant`
 - adds support for `java.time.OffsetDateTime`


.ruby-version
```diff
@@ -1 +1 @@
-jruby-9.4.13.0
+jruby-10.0.5.0
```
mashhurs: I think we need rollback this till the point we support jruby-10 in CI agents. FYI it is not used by platform (Logstash or plugin) but by infrastructure when gem packaging for E2E tests.
I will bring this to team.

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
-## Unreleased
+## 9.5.0
```
mashhurs, line 1: I was thinking to release 9.4.next version since it is a bugfix. WDYT?

src/main/java/co/elastic/logstash/filters/elasticintegration/IngestDuplexMarshaller.java
```diff
@@ -425,10 +426,14 @@ private Timestamp safeTimestampFrom(final Object object) {
...
                 return timestamp;
-            } else if (object instanceof Instant instant) {
+            } else if (object instanceof java.time.Instant instant) {
                 return new Timestamp(instant);
-            } else if (object instanceof ZonedDateTime zonedDateTime) {
-                return new Timestamp(zonedDateTime.toInstant());
+            } else if (object instanceof java.time.chrono.ChronoZonedDateTime<?> chronoZonedDateTime) {
+                return new Timestamp(chronoZonedDateTime.toInstant());
+            } else if (object instanceof java.time.OffsetDateTime offsetDateTime) {
+                return new Timestamp(offsetDateTime.toInstant());
```
mashhurs, line 434: `Instant`, `ChronoZonedDateTime`, `OffsetDateTime`, `ZonedDateTime`, etc.. they all implements [`TemporalAccessor` interface](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/time/temporal/TemporalAccessor.html). WDYT if we make an abstraction here, something like this?
```
} else if (externalObject instanceof TemporalAccessor temporalAccessor) {
            return new Timestamp(Instant.from(temporalAccessor));
}
```

One additional thought is, when using `time.*` interfaces, there might be cases setting `ZoneId`, for example - https://github.com/elastic/integrations/blob/d57619a6e78c67703edd701f971b600b75335ec6/packages/nginx_ingress_controller/data_stream/error/elasticsearch/ingest_pipeline/default.yml#L66-L67
Do we also need to add a `ZoneId` support?

src/main/java/co/elastic/logstash/filters/elasticintegration/IngestDuplexMarshaller.java
```diff
@@ -425,10 +426,14 @@ private Timestamp safeTimestampFrom(final Object object) {
...
+            } else if (object instanceof java.util.Date utilDate) {
```
mashhurs, line 435: ❤️

mashhurs (changes requested):

CHANGELOG.md
yaauie: Yes, by all means let's backport to 9.4 etc. But the branch's `VERSION` already is `9.5.0`.

src/main/java/co/elastic/logstash/filters/elasticintegration/IngestDuplexMarshaller.java
yaauie: While they all implement `TemporalAccessor`, `Instant#from(TemporalAccessor)` will throw a `DateTimeException` for many implementations of `TemporalAccessor` because they don't contain enough information to unambiguously mean a single instant on the timeline.

I specifically selected the java-core implementations that _do_ map to a specific instant.

.ruby-version
yaauie: 🤔 hmm. My local build failed with local LS 9.4 until I did this.

src/main/java/co/elastic/logstash/filters/elasticintegration/IngestDuplexMarshaller.java
yaauie: > Do we also need to add a ZoneId support?

I don't think so, and I would rather not.

At the stage where we're converting the temporary elasticsearch-IngestDocument into a logstash-Event, we don't have enough context to be injecting a zone into any objects that don't happen to have any offset information. This needs to be done _inside_ the ingest pipeline, with the context of the pipeline, and it outside the scope of what this plugin can meaningfully have opinions about.

.ruby-version
yaauie:
```suggestion
jruby-9.4.13.0
```

src/main/java/co/elastic/logstash/filters/elasticintegration/IngestDuplexMarshaller.java
mashhurs: > will throw a DateTimeException for many implementations of TemporalAccessor because they don't contain enough information

Ah right!

Thank you for clarification.

mashhurs (approved):
LGTM!

mashhurs:
I have run E2E with squid package and confirm the fix, merging...

mashhurs:
@mergifyio backport 9.4

PR #467, opened 2026-06-26
Bump jackson dependency

donoghuc:
Thanks for contributing to Logstash! If you haven't already signed our CLA, here's a handy link: https://www.elastic.co/contributor-agreement/

PR #468, opened 2026-06-29
chore: Update changelog and bump version to 9.5.1 for 'Bump jackson dependency (#467)'

donoghuc:


mashhurs (approved):
❤️

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 9.5.1
```
mashhurs, line 1: Ah it looks like this should go under `UNRELEASED` (below)

mashhurs (changes requested):

PR #480, opened 2026-07-20
Set SO_TIMEOUT on IOReactor to prevent indefinite connection hangs

kaisecheng:
Set `IOReactorConfig.setSoTimeout()` on the HTTP async client's IO reactor to match the ES RestClient's default socket timeout (30s). This ensures every NIO `IOSession` has a baseline `SO_TIMEOUT` from creation, independent of the per-request timeout applied by `MainClientExec.generateRequest()`.

### DIFF
--- a/src/main/java/co/elastic/logstash/filters/elasticintegration/ElasticsearchRestClientBuilder.java
+++ b/src/main/java/co/elastic/logstash/filters/elasticintegration/ElasticsearchRestClientBuilder.java
@@ -17,6 +17,7 @@ import org.apache.http.HttpRequestInterceptor;
 import org.apache.http.conn.ssl.NoopHostnameVerifier;
 import org.apache.http.conn.ssl.TrustAllStrategy;
 import org.apache.http.impl.nio.client.HttpAsyncClientBuilder;
+import org.apache.http.impl.nio.reactor.IOReactorConfig;
 import org.apache.http.message.BasicHeader;
 import org.apache.http.protocol.HttpContext;
 import org.apache.http.ssl.SSLContextBuilder;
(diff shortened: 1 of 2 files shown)


kaisecheng:
/pr_prepare release:patch
- Set SO_TIMEOUT on IOReactor to ensure NIO connections have a baseline socket timeout
  - Prevent indefinite hangs when connection pool restore fails to apply the request-level timeout

src/main/java/co/elastic/logstash/filters/elasticintegration/ElasticsearchRestClientBuilder.java
```diff
@@ -191,6 +192,12 @@ RestClient build(final RestClientBuilder restClientBuilder) {
             if (Objects.nonNull(this.userAgentHeaderValue)) {
                 httpClientBuilder.setUserAgent(this.userAgentHeaderValue);
             }
+
+            httpClientBuilder.setDefaultIOReactorConfig(
+                IOReactorConfig.custom()
+                    .setSoTimeout(RestClientBuilder.DEFAULT_SOCKET_TIMEOUT_MILLIS)
```
mashhurs, line 198: I think for now it looks a right to set the timeout but for the long term for the given ES is not responsible, doesn't plugin need to still use the recent pipeline it resolved (w/o invalidating the cache)? - WDYT @yaauie ?

mashhurs (approved):
LGTM!

src/main/java/co/elastic/logstash/filters/elasticintegration/ElasticsearchRestClientBuilder.java
kaisecheng: For more context, the cached value has a maximum TTL of 24 hours. In the support case, the last resolved value kept the pipeline working for 15+ more hours. 

```log
reload-failure(pipeline) { xxx } the existing cached empty value will continue to be available until it expires in ~15h12m38s
```

I’m not sure about the requirements of this plugin, but in general, running a stalled ingest pipeline for that many hours doesn’t sound right to me.

donoghuc:
Will this be backported to all active release tracks?

kaisecheng:
> Will this be backported to all active release tracks?

I plan to do so

donoghuc (approved):
Thanks, I believe this will help the case where a connection is silently closed by an intermediate proxy.

